PR #6, opened 2016-07-13
[twemproxy] adding integration.

truthbk:
Adding twemproxy integration using the SDK testing/scaffolding facilities.


truthbk:
change of heart, this goes in `integrations-core`.

PR #8, opened 2016-07-25
Adds MongoDB

gmmeyer:
This will add the mongo check to the integrations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ install:
 script:
   - bundle exec rake lint
   - rake ci:run[default]
+  - rake ci:run[mongo]
   - bundle exec rake requirements
 
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ install:
 
 script:
   - bundle exec rake lint
-  - rake ci:run[default]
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -40,6 +40,7 @@ install:
 
 script:
   - bundle exec rake lint
+  - rake ci:run[mongo]
   - rake ci:run[default]
```
truthbk: now that the new gem with your fixes has been released, can we put [default] back as the first rake task?

mongo/test_mongo.py
```diff
@@ -0,0 +1,285 @@
...
+        self.assertEquals('Starting Up', self.check.get_state_description(0))
+        self.assertEquals('Recovering', self.check.get_state_description(3))
+
+        # Unknown states:
+        self.assertEquals('UNKNOWN', self.check.get_state_name(500))
+        unknown_desc = self.check.get_state_description(500)
+        self.assertTrue(unknown_desc.find('500') != -1)
+
+@attr(requires='mongo')
+class TestMongo(unittest.TestCase):
```
truthbk: Did this one run as well?

PR #9, opened 2016-07-25
[sdk] improving gitignore.

truthbk:


gmmeyer:
👍

jyee:
Made PR #10 against this to add .ruby-version which is used by rbenv and chruby

PR #11, opened 2016-07-27
adds pylint to test requirements

gmmeyer:
Pylint isn't in the test requirements, which produces (because the command itself is a bit enigmatic) an extremely enigmatic error when you try to run rake. This fixes that by adding pylint to test-requirements.txt.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -2,3 +2,4 @@ nose==1.3.4
 flake8==2.5.1
 mock==1.0.1
 pep8==1.5.7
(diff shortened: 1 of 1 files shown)


truthbk:
I think I put this in my `twemproxy` integration as well, should've put it in its own PR like you did! Much better, thanks! :shipit:

truthbk:
The CI is broken because the `gem`doesn't handle it correctly when there-s no `py` files... we should be good to merge regardless.

PR #10, opened 2016-07-26
Adding rbenv .ruby-version

jyee:
rbenv stores the local ruby version in .ruby-version

PR #15, opened 2016-07-30
Add rbenv ruby version file to gitignore.

jyee:
New PR since the previous PR (#10)  on @truthbk's gitignore didn't make it in time.


irabinovitch:
@truthbk @jyee  can we merge this?

PR #22, opened 2017-01-18
Added Burrow integration

lypnol:


jeremy-lq:
@lypnol has indicated he's willing to be the maintainer for this integration. Can we get a review?

burrow/check.py
```diff
@@ -0,0 +1,86 @@
...
+        if 'status_url' not in instance:
+            raise Exception("Missing 'status_url' in Burrow config")
+
+        url = instance['status_url']
+        connect_timeout = int(instance.get('connect_timeout', 5))
+        receive_timeout = int(instance.get('receive_timeout', 15))
+
+        optional_tags = instance.get('tags', [])
+
+        r = requests.get(url=url+Burrow.LIST_CLUSTERS,
```
olivielpeau: could you replace all references of `Burrow` by `self`? More consistent with the rest of the integrations' codebase

burrow/check.py
```diff
@@ -0,0 +1,86 @@
...
+    LIST_CONSUMERS = '/v2/kafka/{cluster}/consumer'
+    CONSUMER_LAG = '/v2/kafka/{cluster}/consumer/{consumer}/lag'
+
+    @staticmethod
+    def _check_for_error(req):
+        if req.status_code != 200:
+            raise Exception("Got {} error from burrow http endpoint".format(req.status_code))
+
+        res = req.json()
+        if res['error']:
```
olivielpeau: is the `error` field always present in the json object? If it's not you might want to use `if res.get('error'):` instead

burrow/check.py
```diff
@@ -0,0 +1,86 @@
...
+            raise Exception("Got error from burrow http endpoint: {}".format(res['message']))
```
olivielpeau: (same comment as above about the `message` field)

burrow/conf.yaml.example
```diff
@@ -0,0 +1,17 @@
...
+    # tags:
+    #   - optional_tag
+
+    # The (optional) connect_timeout will override the default value, and fail
+    # the check if the time to establish the (TCP) connection exceeds the
+    # connect_timeout value (in seconds)
+    # connect_timeout: 5
+
+    # The (optional) receive_timeout will override the default value, and fail
+    # the check if the time to receive the server status from the Apache server
```
olivielpeau: `Apache` should probably be removed

circle.yml
```diff
@@ -35,6 +35,7 @@ test:
     override:
         - bundle exec rake lint
         - rake ci:run[default]
+        - rake ci:run[burrow]
```
olivielpeau: could you add this line to the `.travis.yml` file instead? We're not using circleCI anymore

burrow/test_burrow.py
```diff
@@ -0,0 +1,30 @@
...
+class TestBurrow(AgentCheckTest):
+    """Basic Test for burrow integration."""
+    CHECK_NAME = 'burrow'
+
+    def test_check(self):
+        """
+        Testing Burrow check.
+        """
+        self.load_check({}, {})
+        self.assertTrue(True)
```
olivielpeau: looks like this test is not finished, let me know if you'd like some pointers on how to write it

burrow/metadata.csv
```diff
@@ -0,0 +1 @@
+kafka.consumer.lag,gaugue,,offset,,kafka consumer lag,-1,burrow,kafka consumer lag
```
olivielpeau: should be `gauge`

Can you also keep the line of column headers?

burrow/test_burrow.py
lypnol: I would prefer to have some indications to write tests :)  
(sorry for the delay...)

jeremy-lq: @lypnol We're working on getting Agent6 shipped now, but if you have specific questions about the test @olivielpeau is happy to assist as time permits. Alternatively, feel free to stop by Community Office hours for real-time assistance (https://github.com/DataDog/dd-agent/wiki/Community-Office-Hours). Thanks again for the contribution!

PR #25, opened 2017-01-30
[ci] fixing/tweaking Travis, Appveyor and CircleCI

truthbk:
Title sums it up.


gmmeyer:
circle failed but I'm not really that worried about that, I turned on appveyor. Can you make a dummy push to this branch to make sure it works?

PR #34, opened 2017-01-31
[apt] adding integration

truthbk:
### What does this PR do?

Adds an apt integration.

Originally written by @ejholmes (https://github.com/DataDog/dd-agent/pull/2275).

### Motivation

This is a custom agent check that will generate metrics and service alerts for apt package updates. It uses the /var/lib/update-notifier/updates-available file to determine what package updates are available (see http://askubuntu.com/questions/49958/how-to-find-the-number-of-packages-needing-update-from-the-command-line).

If there are security related updates available, the apt.updates service check goes critical. If there are non-security related package updates, the apt.updates check warns.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ env:
     - DD_AGENT_BRANCH=master
   matrix:
     - TRAVIS_FLAVOR=default
+    - TRAVIS_FLAVOR=apt FLAVOR_VERSION=latest
     # END OF TRAVIS MATRIX
 
 before_install:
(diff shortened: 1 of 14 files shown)


directionless:
My comments from https://github.com/DataDog/dd-agent/pull/2275

I think using this requires update-notifier, which claims to be a damon, and thus feels very heavy. Thoughts about things like apt-get -s dist-upgrade | grep upgraded or apt-get -s dist-upgrade | grep ^Inst | wc -l

I think it's also exposed in the python apt bindings:
```
import apt
cache = apt.Cache()
cache.upgrade(dist_upgrade=True)
cache.install_count
cache.delete_count
```

PR #42, opened 2017-02-01
[python_rq] adding integration

truthbk:
## What does this PR do?

This PR adds the [python-rq](http://python-rq.org/) agent check. RQ uses Redis as a backing service to store jobs states each time one of them is enqueued. Users may define their own queues and may choose how many workers are assigned to each queue.

Originally written by @palazzem (https://github.com/DataDog/dd-agent/pull/2569)

### Motivation

The `AgentCheck` collects metrics so that developers are aware of the queues usage, while knowing how much their workers are busy. This is critical to decide when it's the time to spawn more workers or to assign more workers to a particular queue.
## Collected metrics
- `python_rq.queue.enqueued`
- `python_rq.queue.in_progress`
- `python_rq.queue.deferred`
- `python_rq.queue.finished`
- `python_rq.queue.failed`
- `python_rq.workers`

Queue metrics are tagged using the queue name (i.e. `queue:high`); workers metric is tagged using the current worker status (i.e. `status:idle`).
## Configuration file

It exposes the `queues` parameter so that users may decide to monitor only the chosen queues. If the value is not present, all queues are monitored.
## Notes
- no external requirements (it uses just `redis`)
- Redis connections are established and stored using the same approach available in the built-in `Redis` check

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ env:
     - DD_AGENT_BRANCH=master
   matrix:
     - TRAVIS_FLAVOR=default
+    - TRAVIS_FLAVOR=python_rq FLAVOR_VERSION=latest
     # END OF TRAVIS MATRIX
 
 before_install:
--- a/circle.yml
+++ b/circle.yml
@@ -32,6 +32,7 @@ test:
     override:
         - bundle exec rake prep_travis_ci
         - rake ci:run[default]
+        - rake ci:run[python_rq]
         - bundle exec rake requirements
     post:
         - if [[ $(docker ps -a -q) ]]; then docker stop $(docker ps -a -q); fi
--- a/python_rq/README.md
+++ b/python_rq/README.md
@@ -1,0 +1,32 @@
+# Python_rq Integration
+
+## Overview
+
+Get metrics from python_rq service in real time to:
+
+* Visualize and monitor python_rq states
+* Be notified about python_rq failovers and events.
+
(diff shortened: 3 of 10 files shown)


palazzem:
@truthbk I think we may close that PR, since it probably require many changes with the new `integrations-extras`. Do you know if there is someone that asked for this integration? if it was "just me", I think we can simply close it.

jeremy-lq:
Based on the comment from @palazzem I'm going to close this PR. If someone is interested in becoming the maintainer I'll reopen.

dantuluri:
I'm interested! Looked over the code and it LGTM, but not sure how much has progressed since 3 years ago.

If RQ won't be supported, however, I'll just write some scripts on my end.

hithwen:
Hi @dantuluri, we can help you get this to a completion stage if you are interested in taking over and becoming a maintainer.

From the top of my head what you need to do is:

1. Update with master
2. Delete  .travis.yml and circle.yml and add it to the .azure_pipelines instead
3. Delete the rake file
4. Add redis to the requirements file ?

If you require an instance of RQ for tests my suggestion is to use docker. We have plenty of examples of that, you can have a look for example to redis tests: https://github.com/DataDog/integrations-core/tree/master/redisdb/tests, in particular the ones marked as integration or e2e.

> If RQ won't be supported, however, I'll just write some scripts on my end.

What do you mean with this?

PR #44, opened 2017-02-01
[network_ext] adding integration

truthbk:
### What does this PR do?

Adds additional networking related metrics from /proc/net/netstat, /proc/net/snmp and /proc/net/udp under the `system.net` and `docker.net` namespaces.

Originally written by @ejholmes (https://github.com/DataDog/dd-agent/pull/2864)

### Motivation

The built-in network.py check doesn't give an accurate representation of TCP connection states, since it only represents the state at the time the metrics were collected. The raw metrics from /proc/net provide a much more granular and accurate view of networking.

For example, `/proc/net/netstat` gives a much more accurate representation for things like number of TCP connections opened, since it's a monotonically increasing counter within the Kernel.
### Additional Notes

There's a similar PR open at https://github.com/DataDog/dd-agent/pull/2720, but it's not Docker aware, which is a necessity for us at Remind. We've also chosen to use monotonic counters instead of rates, since that gives us a more accurate representation of how many events occurred in a time frame.

Hopefully we can combine ideas from both PR's, and get a more releasable version of this into `network.py`.

### Note (cc @ejholmes)
Changes to the core (ie. docker_daemon), and the addition of the network utils has been removed from this PR when compared to the original. If you still wish us to consider that please feel free a specific PR for that in https://github.com/DataDog/dd-agent repo.


irabinovitch:
@ejholmes  Let us know if you need a hand with getting tests passing, etc.  Feel free to stop by our [office hours](https://github.com/DataDog/dd-agent/wiki/Community-Office-Hours) as well if we can help.  Next one is this morning at 10:30 PT / 13:30 EST.

PR #47, opened 2017-02-03
[snmpwalk] adding integration

truthbk:
### What's this PR do?

Provides a wrapper to the snmpwalk binary.

This integration, originally written by @ross here https://gist.github.com/ross/0f6b7cc57bc3f42e93292fbafc6e1194 is a wrapper on snmpwalk and should behave similarly to the original SNMP integration (https://github.com/DataDog/integrations-core/pull/110).

### Motivation

This was written by @ross as a faster alternative to the original integration. It's merely a parser for the seriously quicker snmpwalk binary. So if you're collecting large volumes of metrics by walking, this provides a means to keep your CPU low with a much quicker check. It's not as feature rich, but may help in numerous use-cases.

This integration relies on the `snmpwalk` binary. This is a hard requirement.

### DIFF
--- a/snmpwalk/check.py
+++ b/snmpwalk/check.py
@@ -1,0 +1,196 @@
+# stdlib
+from collections import defaultdict
+from subprocess import check_output, CalledProcessError
+import re
+
+# 3rd party
+
+# project
+from checks.network_checks import NetworkCheck, Status
+
+
+EVENT_TYPE = SOURCE_TYPE_NAME = 'snmpwrap'
+
+class SnmpwalkCheck(NetworkCheck):
+    '''
+    This is a work-alike for checks.d/snmp.py that makes use of snmpwalk for
+    performance reasons. It's much faster than making the calls directly with
+    pysnmp in our use-cases. It is not a 100% reimplementation. It only supports
+    the functionality that we currently use. Other differences include:
+        * regex matching for dynamic tagging w/additional_tags
+        * tags using "enum" values emit the human-readable form instead of the
+          integer.
+    '''
+
+    DEFAULT_RETRIES = 2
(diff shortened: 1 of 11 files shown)


snmpwalk/check.py
```diff
@@ -0,0 +1,214 @@
...
+        hostname = instance.get('metric_host', None)
+
+        # Build up our dataset
+        data = defaultdict(dict)
+        types = {}
+        for metric in metrics:
+            mib = metric['MIB']
+            table = metric['table']
+            cmd = [self.binary, '-c{}'.format(community_string),
+                   '-v2c', '-t', str(timeout), '-r', str(retries)]
```
jescholl: It looks like you're hardcoding this to use snmp v2, but your example config claims to support v3.  Do you have plans to add v3 support?

ross:
I have quite a few updates to the original version of this since it was passed along as a gist a while back. TBH it's a lot of custom stuff that probably doesn't fit/belong in a general solution, but I haven't had the time to look though and figure that out. Will this be merged or will it just live in this branch?

snmpwalk/check.py
truthbk: @jescholl I can definitely see that happen. Enabling the actual `snmpwalk` to make the v3 request should be straight-forward. Wondering how much work the parser might require.

truthbk:
@ross this will be merged eventually. If you feel there's anything general you would like to contribute, it might make sense to open a PR against the active branch `jaime/snmpwrap`, and we can merge that into this one.

Thanks for your awesome work!

truthbk:
@ross this has now been merged into the repository, so feel free to open a new PR if you wish to add new stuff! Thanks for your work! 🙇

PR #28, opened 2017-01-30
[filebeat] adding integration.

truthbk:
### What does this PR do?

Adds a check for filebeat

Originally written by @wk8 (https://github.com/DataDog/dd-agent/pull/3062)

### Motivation

Filebeat (https://github.com/elastic/beats) is a deamon tailing and processing
log files, for example to send them to Logstash.

It maintains a _registry file_, in which it keeps track of which log files it's
currently tailing, and up to what point it has processed them.

This patch adds an agent check that leverages that registry file to gauge
filebat's progress; more specifically, it pushes how many bytes are currently
left to process for filebeat.

This metric would be useful for example to create a monitor alerting when
filebeat falls too far behind.

### Testing

A Testcase is provided here: tests/checks/mock/test_filebeat.py
It contains 6 tests, which provide 100% code coverage on the new check.

PR #48, opened 2017-02-08
Adding a check for filebeat

wk8:
**What does this PR do?**

Adds a check for filebeat

Originally written by @wk8 (DataDog/dd-agent#3062)
And moved to integrations-extras by @truthbk (https://github.com/DataDog/integrations-extras/pull/28)

**Motivation**

Filebeat (https://github.com/elastic/beats) is a deamon tailing and processing
log files, for example to send them to Logstash.

It maintains a registry file, in which it keeps track of which log files it's
currently tailing, and up to what point it has processed them.

This patch adds an agent check that leverages that registry file to gauge
filebat's progress; more specifically, it pushes how many bytes are currently
left to process for filebeat.

This metric would be useful for example to create a monitor alerting when
filebeat falls too far behind.

**Testing**

A Testcase is provided here: tests/checks/mock/test_filebeat.py
It contains 6 tests, which provide 100% code coverage on the new check.


truthbk:
Hi @wk8 thanks for submitting this here! I have started the review and hope to be able to provide some feedback soon! Thanks for this! 👍

wk8:
Thanks @truthbk  :) !

wk8:
@truthbk : have you had a chance to finish the review for this? Thanks!

wk8:
@truthbk : ping?

wk8:
@truthbk @olivielpeau could you please review this? :)

truthbk:
@wk8 my apologies for the delay - this caught me during my move back to Europe, and it completely fell off my radar. I'm not sure how come I missed the notifications though!

filebeat/check.py
```diff
@@ -0,0 +1,59 @@
...
+            raise Exception('An absolute path to a filebeat registry path must be specified')
+
+        registry_contents = self._parse_registry_file(registry_file_path)
+
+        for item in registry_contents.itervalues():
+            self._process_registry_item(item)
+
+    def _parse_registry_file(self, registry_file_path):
+        try:
+            with open(registry_file_path) as registry_file:
```
truthbk: @wk8 not super familiar with `filebeat`, how often does the file get updated? Is it simply a live directory of files to track?

wk8:
@truthbk no worries, thanks :) especially since you were moving back to Spain!

filebeat/test_filebeat.py
```diff
@@ -0,0 +1,94 @@
...
+            with patch.object(os, 'stat') as patched_os_stat:
+                patched_os_stat.side_effect = internal_mock
+                return function(*args, **kwargs)
+        return test_wrapper
+
+    return external_wrapper
+
+
+@attr(requires='filebeat')
+class TestFilebeat(AgentCheckTest):
```
truthbk: You're  a legend for these tests! 🍰

filebeat/metadata.csv
```diff
@@ -0,0 +1 @@
+metric_name,metric_type,interval,unit_name,per_unit_name,description,orientation,integration,short_name
```
truthbk: could you please include the metadata details for `filebeat.registry.unprocessed_bytes` here?

filebeat/test_filebeat.py
```diff
@@ -0,0 +1,94 @@
...
+# project
+from tests.checks.common import AgentCheckTest, Fixtures
+
+
+mocked_file_stats = namedtuple('mocked_file_stats', ['st_size', 'st_ino', 'st_dev'])
+
+
+# allows mocking `os.stat` only for certain paths; for all others it will call
+# the actual function - needed as a number of test helpers do make calls to it
+def with_mocked_os_stat(mocked_paths_and_stats):
```
truthbk: This is fine, but perhaps over-complicated? We're not really every going to stat a real file from what I see in the fixtures, right? Why not just use a full `Mock` and pass it to patch as the `new_callable`. Maybe I'm missing, by all means let me know if I'm missing something.

filebeat/check.py
```diff
@@ -0,0 +1,59 @@
...
+                return json.load(registry_file)
+        except IOError:
+            self.log.warn('No filebeat registry file found at %s' % (registry_file_path, ))
+            return {}
+
+    def _process_registry_item(self, item):
+        source = item['source']
```
truthbk: So, this may be a bit of a concern in terms of the volume of contexts that may be created. Are all possible `source`s a large set of files? Is it constrained? Are these `sources` short-lived?

filebeat/check.py
wk8: Nope, you tell filebeat which files you want to tail, and it keeps track in its registry of where it is at for each of them.

filebeat/check.py
wk8: I'm not sure I understand what you mean about the volume of contexts, I'm sorry?

As for sources, no, there might be a large number; but filebeat is already having to do a decent amount of work for each of those, the overhead added in the agent for each is neglectable in comparison.

filebeat/metadata.csv
wk8: Sure thing :)

filebeat/test_filebeat.py
wk8: I'm not sure I understand what you're suggesting, sorry? :)

The reason why this is somewhat complicated is that, as mentioned in the comment just above, a number of test helpers from `AgentCheckTest` _do_ need to stat real files, and so mocking `os.stat` altogether is not an option

filebeat/test_filebeat.py
wk8: Thanks! :)

wk8:
@truthbk : thanks for the review, amended/responsed.

Not sure what you mean about contexts though? :)

Thanks!

filebeat/check.py
truthbk: So the set of files therein is static presumably. That's good.

filebeat/check.py
truthbk: No, no... I understand the heavy lifting is already done by filebeat. My concern here is that we might create an unbounded amount of "contexts" (a distinct set of tags for a metric)... That would be undesireable. The question I asked above about the files contained in the registry, and if it was a live registry, was also geared towards trying to understand that. I take the "source" is typically the filename, correct (from my limited research it looks that way)?

filebeat/test_filebeat.py
truthbk: Gotcha. We can keep this as is 👍

truthbk:
@wk8 added some more feedback... Maybe with the definition I gave you in one of the comments you can get a better sense into what was meant by "context". I'd love to hear your thoughts.

filebeat/check.py
wk8: Indeed it is. Typically, filebeat won't track more than a handful of files per box

filebeat/check.py
wk8: Yup, "source"s are the names of the files that filebeat tails, typically a handful

wk8:
@truthbk Thanks for the review, answered your outstanding questions :)

wk8:
@truthbk : sure I'm happy being the maintainer :) please let me know if there's anything else I should do right away. Thanks!

jeremy-lq:
Thanks, @wk8. We'll get this merged ASAP now that you've confirmed.

jeremy-lq:
Thanks again for the contribution, @wk8! Everything looks good, so once the manifest is updated we can merge.

wk8:
@jeremy-lq @truthbk thanks updated the manifest :)

jeremy-lq:
Thanks @wk8! @truthbk, do we need anything else to get this merged?

filebeat/check.py
```diff
@@ -0,0 +1,59 @@
+# (C) Datadog, Inc. 2010-2016
+# All rights reserved
+# Licensed under Simplified BSD License (see LICENSE)
+
+# stdlib
+
+# 3rd party
+import json
```
remh: nitpick but this should be using simplejson instead (way better perf)

wk8: Sure :)

filebeat/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "maintainer": "wk8",
```
truthbk: would you be comfortable adding an email address here?

snmpwalk/metadata.csv
```diff
@@ -1 +1,2 @@
 metric_name,metric_type,interval,unit_name,per_unit_name,description,orientation,integration,short_name
+filebeat.registry.unprocessed_bytes,gauge,"",byte,"",Number of bytes not processed by filebeat yet.,0,filebeat,""
```
truthbk: Hi @wk8 you somehow added the metric to the `snmpwalk` metadata instead of the `filebeat`. Could you please update this?

filebeat/metadata.csv
wk8: (done)

wk8:
@truthbk addressed your latest comments :)

PR #51, opened 2017-03-15
Neo4j Initial Integration

davidfauth:
Initial Neo4j Datadog Files


gmmeyer:
Hey @davidfauth! Thank you very much for your contribution!

I'm going to review this fully, however can you fix the linting errors while I look over this more fully? 😃

davidfauth:
Yes.

neo4j/test_neo4j.py
```diff
@@ -0,0 +1,158 @@
...
+        'readonly',
+        'storecreationdate',
+        'storelogversion',
+        'kernelstarttime',
+        'lastcommittedtxid',
+        'peaknumberofconcurrenttransactions',
+        'numberofrolledbacktransactions',
+        'numberofopentransactions',
+        'numberofopenedtransactions',
+        'numberofcommittedtransactions',
```
irabinovitch: I'm not sure the number prefix adds much value here. Would it be more readable without that?

neo4j/metadata.csv
```diff
@@ -0,0 +1,37 @@
...
+fileunmappings,gauge,,unit,,The total number of file unmappings.,0,neo4j,File Unmappings
+flushes,gauge,,unit,,The total number of flushes.,0,neo4j,Flushes
+ha.pull_interval,gauge,,unit,,The HA Pull Time Interval,0,neo4j,Pull Interval
+kernelstarttime,gauge,,second,,Time when the kernel started,-1,neo4j,Kernel Start Time
+kernelversion,gauge,,unit,,The kernel version,0,neo4j,Kernel Version
+lastcommittedtxid,gauge,,unit,,The last committed transaction id.,0,neo4j,Last Committed Transaction
+locks,gauge,,unit,,The total number of locks.,0,neo4j,Number of Locks
+logicallogsize,gauge,,unit,,The logical log size setting.,0,neo4j,Logical Log Size
+memorypools,gauge,,unit,,The size of memory pools.,0,neo4j,Memory Pools
+nodestoresize,gauge,,unit,,The size of the node store.,0,neo4j,Node Store Size
```
irabinovitch: would neo4j.nodestore.size be more readable?

neo4j/metadata.csv
```diff
@@ -0,0 +1,37 @@
...
+numberofaverteddeadlocks,gauge,,unit,,The total number of averted deadlocks.,-1,neo4j,Number of Averted Deadlocks
+numberofcommittedtransactions,gauge,,unit,,The total number of committed transactions.,0,neo4j,Number of Committed Transaction
+numberofnodeidsinuse,gauge,,unit,,The total number of node ids in use.,-1,neo4j,Number of Nodes in Use
+numberofopenedtransactions,gauge,,unit,,The total number of opened transactions.,-1,neo4j,Number of Opened Transactions
```
irabinovitch: I'd recommend dropping the numberof prefix here to make this more readable.

neo4j/check.py
```diff
@@ -0,0 +1,148 @@
...
+    def timeout_event(self, url, timeout, aggregation_key):
+        self.event({
+            'timestamp': int(time.time()),
+            'event_type': 'http_check',
+            'msg_title': 'URL timeout',
+            'msg_text': '%s timed out after %s seconds.' % (url, timeout),
+            'aggregation_key': aggregation_key
+        })
+
+    def status_code_event(self, url, r, aggregation_key):
```
irabinovitch: It looks like this might not be in use anywhere. Should we remove it?

neo4j/neo4j.yaml
```diff
@@ -0,0 +1,12 @@
+init_config:
```
irabinovitch: This file looks to have been committed by mistake. We should only include conf.yaml.example.

neo4j/conf.yaml.example
```diff
@@ -0,0 +1,12 @@
+init_config:
+    default_timeout: 5
+
+instances:
+    -   url: http://ec2-54-84-9-93.compute-1.amazonaws.com:7474/
+        port: 7474
+        server: http://ec2-54-84-9-93.compute-1.amazonaws.com
```
irabinovitch: Should this be neo4j_server? Could  we change the example from this EC2 host name to something that is more clearly an example? Finally, it looks like server and url might be duplicates?

neo4j/requirements.txt
```diff
@@ -0,0 +1,6 @@
+# integration pip requirements
+time
```
irabinovitch: Time is part of the standard library so does not need to be included as a dependency.

neo4j/requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+requests
+hashlib
```
irabinovitch: hashlib is part of the standard library so does not need to be included as a dependency.

neo4j/requirements.txt
```diff
@@ -0,0 +1,6 @@
+# integration pip requirements
+time
+requests
```
irabinovitch: We should be able to skip requests and simplejson as they are already included in the core agent.  It would be better to avoid including them again here so that we aren't requiring differing versions that may conflict.

https://github.com/DataDog/dd-agent/blob/master/requirements.txt#L27

neo4j/requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+hashlib
+httplib2
```
irabinovitch: Could we pin httplib2 to a specific version? This well help ensure we are building with the version you have tested and avoid builds breaking later.

neo4j/check.py
```diff
@@ -0,0 +1,148 @@
...
+      
+
+# Neo specific
+# Create payload using built-in Neo4j queryJmx stored procedure
+        try:
+            payload = {"statements" : [{"statement" : "CALL dbms.queryJmx('org.neo4j:*') yield attributes with  keys(attributes) as k, attributes unwind k as row return row, attributes[row]['value'];"}]}
+            headers_sent = {'Content-Type':'application/json','Authorization':'Basic ' + b64Val + '','Content-Type':'application/json'}
+            r = requests.post(checkURL, data=json.dumps(payload),headers=headers_sent)
+
+        except (socket.timeout, socket.error, HttpLib2Error) as e:
```
irabinovitch: Is this required? The check relies on request which does not make use of HttpLib2, as such I don't believe we'd ever hit this exception.  Dropping this would also allow us to drop httplib2 as a dependency.

neo4j/test_neo4j.py
```diff
@@ -0,0 +1,158 @@
...
+    }]
+
+
+    CONNECTION_FAILURE = [{
+        'server': 'localhost',
+        'user': 'unknown',
+        'pass': 'dog',
+    }]
+
+    NEO4J_VARS = [
```
davidfauth: Updated with neo4j. prefix.

neo4j/metadata.csv
```diff
@@ -0,0 +1,37 @@
+metric_name,metric_type,interval,unit_name,per_unit_name,description,orientation,integration,short_name
```
davidfauth: Prefixed metrics with neo4j. to make them more readable.

neo4j/neo4j.yaml
```diff
@@ -0,0 +1,12 @@
...
+    default_timeout: 5
+
+instances:
+    -   neo4j_url: http://ec2-54-85-23-10.compute-1.amazonaws.com
```
davidfauth: Removed neo4j.yaml.

neo4j/check.py
davidfauth: Removed.

neo4j/check.py
```diff
@@ -0,0 +1,148 @@
...
+        })
+
+    def _get_config(self, instance):
+        self.host = instance.get('neo4j_url', '')
+        self.port = int(instance.get('port', 7474))
+        user = instance.get('user', '')
+        password = str(instance.get('password', ''))
+        connect_timeout = instance.get('connect_timeout', None)
+        server_name = instance.get('server_name','')
+        version = instance.get('neo4j_version', '3.1')
```
davidfauth: Still need version. Will look into parsing it from a metric.

neo4j/neo4j.yaml
davidfauth: removed.

neo4j/check.py
gmmeyer: You could probably get it from the service root, it seems to be relatively stable across versions in my quick look at it.

https://neo4j.com/docs/rest-docs/current/#rest-api-get-service-root
https://neo4j.com/docs/2.3.9/rest-api-service-root.html
http://neo4j.com/docs/2.1.8/rest-api-service-root.html
http://neo4j.com/docs/1.8.3/rest-api-service-root.html

neo4j/ci/neo4j.rake
```diff
@@ -0,0 +1,80 @@
...
+def neo4j_version
+  ENV['FLAVOR_VERSION'] || 'latest'
+end
+
+def neo4j_rootdir
+  "#{ENV['INTEGRATIONS_DIR']}/data_#{neo4j_version}"
+end
+
+container_name = 'dd-test-neo4j'
+container_port = 13306
```
irabinovitch: Looks like this isn't used?

neo4j/ci/neo4j.rake
```diff
@@ -0,0 +1,80 @@
...
+      install_requirements("--cache-dir #{ENV['PIP_CACHE']}",
+                           "#{ENV['VOLATILE_DIR']}/ci.log", use_venv)
+      sh %(docker run --name #{container_name} --publish=7474:7474 --publish=7687:7687 --volume=$HOME/neo4j/data:/data --volume=$HOME/neo4j/logs:/logs neo4j:3.1.1 )
+    end
+
+    task before_script: ['ci:common:before_script'] do
+      Wait.for 7474
+      count = 0
+      logs = `docker logs dd-test-neo4j 2>&1`
+      puts "Waiting for Neo4j to come up"
```
irabinovitch: NIT but will make linting pass:  Prefer single-quoted strings when you don't need string interpolation or special symbols.

neo4j/ci/neo4j.rake
```diff
@@ -0,0 +1,80 @@
...
+      until count == 60 or logs.include? 'Remote interface available at'
```
irabinovitch: Rubocop is asking that we use || instead of OR here.

neo4j/ci/neo4j.rake
```diff
@@ -0,0 +1,80 @@
...
+        sleep_for 2
+        logs = `docker logs dd-test-neo4j 2>&1`
+        count += 1
+      end
+      if logs.include? 'Remote interface available at'
```
irabinovitch: From rubocop and linting:

C: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.

neo4j/ci/neo4j.rake
```diff
@@ -0,0 +1,80 @@
...
+        puts "Neo4j is up!"
```
irabinovitch: NIT but will make rubocop pass: Prefer single-quoted strings when you don't need string interpolation or special symbols.

davidfauth:
Linted check.py

gmmeyer:
@davidfauth there's a syntax error in the rakefile that's causing the tests to fail

neo4j/check.py
davidfauth: Added in ability to get version programmatically.
Removed from conf.yaml example.

davidfauth:
Fixed the syntax errors in the rake file @gmmeyer

neo4j/requirements.txt
```diff
@@ -0,0 +1,3 @@
+# integration pip requirements
+hashlib
+httplib2==0.9.2
```
truthbk: We currently have a version mismatch here, this is already defined in `dd-agent`, see error logs in the CI for more details:
```
There's a version mismatch with httplib2 ==0.9
```

Do you need this to be 0.9.2? Also, we should pin hashlib to a specific version (not sure if we're using it elsewhere), and on top of it I think hashlib is having problems installing/building in Travis 😞 

I'll try to take a look see if I can spot what the issue is.

neo4j/ci/neo4j.rake
```diff
@@ -0,0 +1,81 @@
+require 'ci/common'
```
irabinovitch: @davidfauth Checking back in on this one. Thanks.

irabinovitch:
Looks like the only remaining test failures are the linting errors we mentioned earlier. If we can address this I think we're likely in a good spot for merging:

neo4j/ci/neo4j.rake:32:25: C: Use || instead of or.
      until count == 60 or logs.include? 'Remote interface available at'
                        ^^
neo4j/ci/neo4j.rake:37:7: C: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
      if logs.include? 'Remote interface available at'
      ^^
neo4j/ci/neo4j.rake:41:1: C: Extra blank line detected.
neo4j/ci/neo4j.rake:41:1: C: Extra empty line detected at block body end.
neo4j/ci/neo4j.rake:79:4: C: Final newline missing.

leifg:
Is there any way to help out in order to get this merged?

jeremy-lq:
@leifg Would you be interested in becoming the maintainer for this integration?

davidfauth:
I think this is the only one question I have left. Apologize for the delay.

neo4j/ci/neo4j.rake:37:7: C: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
if logs.include? 'Remote interface available at'
^^

neo4j/ci/neo4j.rake
```diff
@@ -0,0 +1,78 @@
...
+                           "#{ENV['VOLATILE_DIR']}/ci.log", use_venv)
+      sh %(docker run --name #{container_name} --publish=7474:7474 --publish=7687:7687 --volume=$HOME/neo4j/data:/data --volume=$HOME/neo4j/logs:/logs neo4j:3.1.1 )
+    end
+
+    task before_script: ['ci:common:before_script'] do
+      Wait.for 7474
+      count = 0
+      logs = `docker logs dd-test-neo4j 2>&1`
+      puts 'Waiting for Neo4j to come up'
+      until count == 60 || logs.include? 'Remote interface available at'
```
gmmeyer: you need to put a do after this until statement

```
until count == 60 || logs.include? 'Remote interface available at' do
```

gmmeyer:
@davidfauth I don't like that rubocop setting and I think we'll likely change it. It's an annoying setting. What it's saying is that they want it to be a one liner:

```
logs.include?('Remote interface available at') && puts 'Neo4j is up!'

gmmeyer:
You need to do two other small thing as well before we can approve this. Right now your tests won't run. You need to add your tests here:

gmmeyer:
You need to do two other small thing as well before we can approve this. Right now your tests won't run. You need to add your tests here: 

https://github.com/DataDog/integrations-extras/blob/master/circle.yml#L44

And to the travis flavors:

https://github.com/DataDog/integrations-extras/blob/master/.travis.yml#L42

Otherwise, once the tests pass and the linting passes, I think this will be able to be merged.

irabinovitch:
@davidfauth We'd love to knock this out. Have any bandwidth this month?

neo4j/check.py
davidfauth: I think we can remove httplib2.

gmmeyer:
I'm ready to merge this. @davidfauth are you good with this?

davidfauth:
@gmmeyer I'm good with it.

gmmeyer:
🎉

PR #52, opened 2017-03-15
Linting, Requirements and gitignore should be the same as core

gmmeyer:
These files had all fallen behind core, this brings them up to speed.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,5 +1,6 @@
-nose==1.3.4
 flake8==2.5.1
 mock==2.0.0
 pep8==1.5.7
 pylint==1.6.4
+nose==1.3.4
+psycopg2==2.6.2 # required by postgres tests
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -3,3 +3,4 @@ simplejson==3.6.5
 tornado==3.2.2
 pyyaml==3.11
 docker-py==1.10.6
+kazoo==2.2.1
```
truthbk: is kazoo a requirement here?

requirements-test.txt
```diff
@@ -1,5 +1,6 @@
-nose==1.3.4
 flake8==2.5.1
 mock==2.0.0
 pep8==1.5.7
 pylint==1.6.4
+nose==1.3.4
+psycopg2==2.6.2 # required by postgres tests
```
truthbk: is psycopg2 a requirement here?

gmmeyer: No, probably not, I just copied the two req files over

PR #53, opened 2017-03-15
Updates Travis and Appveyor Config from Core

gmmeyer:
Some changes in Core have been made without copying them here. This will make them equivalent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ env:
     # END OF TRAVIS MATRIX
 
 before_install:
-  - sudo apt-get update ; sudo apt-get install -y curl apt-transport-https git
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -39,7 +39,7 @@ env:
     # END OF TRAVIS MATRIX
 
 before_install:
-  - sudo apt-get update ; sudo apt-get install -y curl apt-transport-https git
+  - sudo apt-get update ; sudo apt-get install -y curl apt-transport-https git netcat net-tools libpq-dev
```
truthbk: Not sure we need to install these, the truth is there's no harm done, but I don't think these are required. I think it's fine if we diverge on the packages installed.

gmmeyer: You're probably right, I can remove them, let me make sure there's no other reasonable use for them.

PR #55, opened 2017-03-29
[storm] added apache storm 1.x.x metric collection

platinummonkey:
# What does this PR do?

Adds additional Apache Storm cluster, supervisor and topology related metrics from 1 or more configured Apache Storm cluster endpoints.

# Motivation

There is currently no built-in Apache Storm metric collection for the Datadog agent. Looking to contribute back some of metric collection techniques we've developed. 

The Storm UI displays a lot of these metrics, which are also available via the Apache Storm UI REST API: http://storm.apache.org/releases/1.0.3/STORM-UI-REST-API.html

## Additional Notes

Currently this integration does not include authentication, but would be a good future improvement for more secured installations or remote host monitoring. 

Includes:

- Storm Cluster Summary metrics
- Storm Nimbus Summary metrics
- Storm Supervisor Summary metrics
- Storm Topology Summary metrics
- Storm Topology metrics

Does not include currently:

- User Authentication
- Bolt/Spout Component Stats (ie, inputStream/outputStream stats)
- Backwards compatibility for pre 1.x.x line of Apache Storm.


platinummonkey:
bump: review / feedback?

platinummonkey:
Will work with internal teams for merging :) Work at Datadog now :)

jeremy-lq:
Thanks for the update, Cody.

jeremy-lq:
@hush-hush @olivielpeau Can we get a review on this and work toward it being merged? Thanks.

platinummonkey:
adding support for new 1.1.x metrics (with backwards compatibility) and refactoring the tests a little bit. Should see another push here by end of weekend.

platinummonkey:
Ready for review (includes 1.1.x metrics, and refactored a bit for clarity and documentation).

storm/check.py
```diff
@@ -0,0 +1,688 @@
...
+            self.http_prefix = 'http://'
+        self.nimbus_server = instance.get('server', self.init_config.get('server', 'localhost:9005'))
+        self.environment_name = instance.get('environment', self.init_config.get('environment', 'dev'))
+        self.additional_tags = instance.get('tags', [])
+        self.excluded_topologies = instance.get('excluded', [])
+        intervals = instance.get('intervals', self.init_config.get('intervals', [60]))
+        self.intervals = []
+        if isinstance(intervals, (list, tuple)):
+            for interval in intervals:
+                if isinstance(interval, (list, tuple)):
```
hush-hush: use the `extend` method instead of looping through.

storm/check.py
```diff
@@ -0,0 +1,688 @@
...
+            self.http_prefix = 'https://'
+        else:
+            self.http_prefix = 'http://'
+        self.nimbus_server = instance.get('server', self.init_config.get('server', 'localhost:9005'))
+        self.environment_name = instance.get('environment', self.init_config.get('environment', 'dev'))
+        self.additional_tags = instance.get('tags', [])
+        self.excluded_topologies = instance.get('excluded', [])
+        intervals = instance.get('intervals', self.init_config.get('intervals', [60]))
+        self.intervals = []
+        if isinstance(intervals, (list, tuple)):
```
hush-hush: You don't need to introspect `intervals` type and convert them. We can state in the configuration we're expecting a list of int and raise an error if it's not the case.

storm/check.py
```diff
@@ -0,0 +1,688 @@
...
+
+        :param instance: Agent instance.
+        :return: None
+        """
+        # Setup
+        self.update_from_config(instance)
+
+        # Cluster Stats
+        cluster_stats = self.get_storm_cluster_summary()
+        for k, v in self.process_cluster_stats(self.environment_name, cluster_stats).items():
```
hush-hush: Each `process_*` function could be in charge of sending its metrics instead of returning a dict and looping through it. This would also avoid the need for `_gt` to return a tuple with the value and the tags in it.

storm/check.py
```diff
@@ -0,0 +1,688 @@
...
+                    results = self.process_topology_stats(topology_stats=stats, interval=interval)
+                    if len(results) > 0:
+                        interval_results.append(results)
+                metric_stats = self.get_topology_metrics(topology_id=topology_id)
+                metric_results = self.process_topology_metrics(topology_name, metric_stats)
+
+                topology_status = _g(stats, 'unknown', None, 'status').upper()
+                if topology_status != 'ACTIVE':
+                    check_status = AgentCheck.CRITICAL
+                    self.service_check(
```
hush-hush: The `self.service_check` could be put outside the is/else to avoid duplicating it at line 624 (the `message` could use `topology_status` in every case).

storm/check.py
```diff
@@ -0,0 +1,688 @@
...
+                    check_status = AgentCheck.OK
+                    self.service_check(
+                        'topology-check.{}'.format(topology_name),
+                        status=check_status,
+                        message='{} topology is active'.format(topology_name),
+                        tags=['env:{}'.format(self.environment_name),
+                              'environment:{}'.format(self.environment_name)] + self.additional_tags
+                    )
+
+                if len(interval_results) > 0:
```
hush-hush: Here too, `process_topology_stats` should report its metrics. That would remove the following code.

storm/ci/storm.rake
```diff
@@ -0,0 +1,65 @@
...
+  "#{ENV['INTEGRATIONS_DIR']}/storm_#{storm_version}"
+end
+
+namespace :ci do
+  namespace :storm do |flavor|
+    task before_install: ['ci:common:before_install']
+
+    task install: ['ci:common:install'] do
+      use_venv = in_venv
+      install_requirements('storm/requirements.txt',
```
hush-hush: This is already done for every check in the main rake task

storm/ci/storm.rake
```diff
@@ -0,0 +1,65 @@
...
+                           "--cache-dir #{ENV['PIP_CACHE']}",
+                           "#{ENV['VOLATILE_DIR']}/ci.log",
+                           use_venv)
+      # sample docker usage
+      # sh %(docker create -p XXX:YYY --name storm source/storm:storm_version)
```
hush-hush: Could we run docker in the CI task ?

storm/conf.yaml.example
```diff
@@ -0,0 +1,33 @@
+init_config:
+  # Default environment to use for all instances
+  # environment: production
+  #
+  # Default Storm API endpoint (typically this is the UI server on the UI port)
+  #   ex. if storm ui is running on the localhost on port 9005
+  # server: localhost:9005
+  #
+  # Use HTTPS by default
```
hush-hush: we could remove that setting and let the user set it through the server (ie: `server: http://localhost:9005`)

hush-hush:
Hi @platinummonkey,

I added a few comments, don't hesitate to ping me if you have any questions.

storm/check.py
```diff
@@ -0,0 +1,688 @@
...
+    Apache Storm 1.x.x Topology Execution Stats
+    """
+
+    nimbus_server = 'localhost:9005'
+    http_prefix = 'http://'
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+
+    def get_library_versions(self):
```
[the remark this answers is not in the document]
platinummonkey: No idea! if not I'll just remove it, the default is `raise NotImplementedError` which seems to be fine.

platinummonkey:
thanks @hush-hush ! I'll apply these comments and update shortly :)

storm/check.py
```diff
@@ -0,0 +1,573 @@
...
+    return _g(stat_map, default, _float, *components)
+
+
+class StormCheck(AgentCheck):
+    """
+    Apache Storm 1.x.x Topology Execution Stats
+    """
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        self.nimbus_server = 'http://localhost:9005'
```
hush-hush: Having `'http://localhost:9005'` as a class constant could be helpful.

storm/check.py
```diff
@@ -0,0 +1,573 @@
...
+        self.additional_tags = []
+        self.excluded_topologies = []
+        self.environment_name = 'dev'
```
hush-hush: Same for `'dev'`

storm/check.py
```diff
@@ -0,0 +1,573 @@
...
+        :type environment: str
+        :param cluster_stats: Cluster stats response
+        :type cluster_stats: dict
+        :return: Extracted cluster stats metrics
+        :rtype: dict
+        """
+        if len(cluster_stats) == 0:
+            return {}
+        else:
+            version = _g(cluster_stats, 'unknown', None, 'version').replace(' ', '_').lower()
```
hush-hush: I think using `_g` here is unnecessary and should be replaced by:

```python
cluster_stats.get('version', 'unknown')
```
This make it easier to read, a little better performance-wise and more pythonic for contributors.
We should do the same for every other use of `_g` where the function is `None` and `components` is a single string.

storm/check.py
```diff
@@ -0,0 +1,573 @@
...
+    """
+    return _g(stat_map, default, _float, *components)
+
+
+class StormCheck(AgentCheck):
+    """
+    Apache Storm 1.x.x Topology Execution Stats
+    """
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
```
hush-hush: I don't think it's necessary to initialized field here since we will override them in `update_from_config` at each `check` call. We could remove the `__init__` method I guess.

storm/check.py
```diff
@@ -0,0 +1,573 @@
...
+        all_tags = tags + [
+            'env:{}'.format(self.environment_name),
+            'environment:{}'.format(self.environment_name)] + additional_tags
+        self.histogram(
+            metric=metric,
+            value=value,
+            tags=all_tags
+        )
+
+    def update_from_config(self, instance):
```
hush-hush: The default values for `nimbus_server`, `environment_name` and `intervals` should be class constants.

storm/check.py
```diff
@@ -0,0 +1,573 @@
...
+                    self.process_topology_stats(topology_stats=stats, interval=interval)
+                    metric_stats = self.get_topology_metrics(topology_id=topology_id, interval=interval)
+                    self.process_topology_metrics(topology_name, metric_stats, interval=interval)
+
+                    # only report this once.
+                    if topology_status is None and _g(stats, 'unknown', None, 'status') is not None:
+                        topology_status = _g(stats, 'unknown', None, 'status').upper()
+                        if topology_status != 'ACTIVE':
+                            check_status = AgentCheck.CRITICAL
+                            self.service_check(
```
hush-hush: I still think the 2 calls to `self.service_check` should be only one and the if/else should only set the value of `check_status`.

storm/conf.yaml.example
```diff
@@ -0,0 +1,27 @@
+init_config:
+  # Default environment to use for all instances
+  # environment: production
+  #
+  # Default Storm API endpoint (typically this is the UI server on the UI port)
+  #   ex. if storm ui is running on the localhost on port 9005
+  # server: http://localhost:9005
```
hush-hush: Is it interesting for the user to have multiple instances pointing to `localhost:9005` (not an expert on storm) ? If not lets remove this from `init_config` and force it to be set on each `instance`.

storm/ci/storm.rake
```diff
@@ -0,0 +1,60 @@
...
+  "#{ENV['INTEGRATIONS_DIR']}/storm_#{storm_version}"
+end
+
+namespace :ci do
+  namespace :storm do |flavor|
+    task before_install: ['ci:common:before_install']
+
+    task install: ['ci:common:install'] do
+      # sample docker usage
+      # sh %(docker create -p XXX:YYY --name storm source/storm:storm_version)
```
hush-hush: Could we run the docker here ? Will the test pass in the CI if we don't ?

platinummonkey: Yes, was working on that last night, still proving out the mini-cluster setup + build the topology and deploy it, then writing a dedicated integration check for ci-only.  Will update the README for contributors as well.  Does our CI support docker compose? would simplify into a single command and make this more local developer friendly to avoid drift.

storm/conf.yaml.example
platinummonkey: agreed, will remove from defaults.

storm/ci/storm.rake
hush-hush: Docker compose should work (it's used in the varnish CI for integration-core)

storm/check.py
```diff
@@ -205,11 +269,15 @@ def process_cluster_stats(self, environment, cluster_stats):
...
         :rtype: dict
         """
-        if len(cluster_stats) == 0:
-            return {}
-        else:
-            version = _g(cluster_stats, 'unknown', None, 'version').replace(' ', '_').lower()
-            tags = ['stormClusterEnvironment:{}'.format(environment), 'stormVersion:{}'.format(version)]
+        if len(cluster_stats) >= 0:
+            version = _get_string(cluster_stats,
+                                  _get_string(cluster_stats, 'unknown', 'stormVersion'),
```
hush-hush: Can stormVersion be something else than a string ? Do we have to use `_get_string` ?

storm/ci/storm.rake
```diff
@@ -31,6 +31,8 @@ namespace :ci do
       # Deploy the basic WordCountTopology we created earlier.
       sh %(docker run --link storm-nimbus:nimbus -it --rm -v $(pwd)/topology.jar:/topology.jar storm:1.1 storm jar \
         /topology.jar org.apache.storm.starter.WordCountTopology topology)
+      # Wait again for slow topology starts.
+      sh %(sleep 60)
```
hush-hush: We use use docker `healthcheck` or some command to check the state of the topology, if possible instead of a sleep. 
But lets do that in another PR to improve the CI (this one is big enough).

PR #63, opened 2017-05-16
Fixes travis.yaml

gmmeyer:
Travis.yaml was calling snmpwalk with rake[snmpwalk]. It should have been using travis flavor instead.


gmmeyer:
I noticed it after I asked someone to fix their travis.yaml and they did it in a weird way. I wouldn't have fixed it but, for some reason, the tests weren't running

PR #64, opened 2017-05-19
[Traefik] Add traefik check

hkaj:
# What does this PR do?

Adds a [Traefik](https://traefik.io/) check.

# Motivation

- we received a few requests for it
- we want to improve our coverage of the cloud native ecosystem
- the new `prometheus_check` needs a few checks using it so we can see what's to improve
- I use Traefik, love it, and need metrics for it :)


timoreimann:
hey @hkaj,

one of the Traefik maintainers here. :)

First of all: Even though I don't use DataDog, it's great to see you're providing better integration with Traefik. Appreciated!

Skimming over the PR, I saw that you apparently [maintain a mapping of Prometheus metrics used in Traefik](https://github.com/DataDog/integrations-extras/pull/64/files#diff-9a41570ca95801d7da0aedd667152742R15). Out of curiosity, is there any plan on how to stay in sync with Traefik's metrics from your end?

FWIW, there's PR containous/traefik#1504 in-flight right now to add metrics for the retry counter, and we are probably going to have a few more over time.

Thanks again!

hkaj:
Hey @timoreimann 

New metrics will need to be added manually to the check, and these updates will be provided on a best-effort basis. The community can handle it, or create issues to notify us/other users about new metrics to add. This is already the process for kube-state-metrics for example.

We don't want to collect all metrics by default because the prometheus format is quite different from the datadog one, so we need to be able to transform metrics to something that better fits our model. And we can't do it retroactively because it would change metrics that people could already be using for alerting (and thus triggering un-wanted alerts).

Does that make sense?

timoreimann:
@hkaj makes absolute sense. 👍 Thanks for the explanation.

hkaj:
that took me long enough to finish it... Can I get a review on this one @truthbk ?

jeremy-lq:
Spoke with Haissam. He still plans to have this merged, but needs a bit of additional time. Removing the stale tag.

Vlaaaaaaad:
I'd love to see status count( how many 20x, 30x, 40x, 50x) added to this.

kedare:
@jeremy-lq any update ? :)
Can we help in any way ?

jeremy-lq:
@hkaj is this ready for a review from @truthbk or do you have additional commits to push?

PR #67, opened 2017-06-27
aerospike integration fixed

dmotylev:
### What does this PR do?

Adds an integration check for Aerospike. Originally written by @joshk0 (DataDog/dd-agent#2219)

### Motivation

The previous work has not The working integration for Aerospike is required for our project.

### Testing Guidelines

The plugin follows Testing Guide from DD.

### Versioning

Changes in manifest and/or changelog are not required, as previous work has not been finished.

### DIFF
--- a/aerospike/check.py
+++ b/aerospike/check.py
@@ -83,8 +83,27 @@ class AerospikeCheck(AgentCheck):
             self._process_datum(event_type, key, value, tags)
 
     def _process_datum(self, event_type, key, val, tags={}):
+        datatype = 'event'
+
         if val.isdigit():
+            datatype = 'gauge'
+        elif val.lower() in ('true', 'on', 'enable', 'enabled'): # boolean : true
+            val = 1
+            datatype = 'gauge'
+        elif val.lower() in ('false', 'off', 'disable', 'disabled'): # boolean : false
+            val = 0
+            datatype = 'gauge'
+        else:
+            try:
+                float(val)
+                datatype = 'gauge'
+            except ValueError:
+                datatype = 'event'
+
+        if datatype == 'gauge':
             self.gauge(self._make_key(event_type, key), val, tags=tags)
+        #elif
+        #    self.event(self._make_key(event_type, key), val, tags=tags)
 
     @staticmethod
     def _make_key(event_type, n):
--- a/aerospike/check.py
+++ b/aerospike/check.py
@@ -16,8 +16,21 @@ SERVICE_CHECK_NAME = '%s.cluster_up' % SOURCE_TYPE_NAME
 CLUSTER_EVENT_TYPE = SOURCE_TYPE_NAME
 NAMESPACE_EVENT_TYPE = '%s.namespace' % SOURCE_TYPE_NAME
 NAMESPACE_TPS_EVENT_TYPE = '%s.namespace.tps' % SOURCE_TYPE_NAME
+SINDEX_EVENT_TYPE = '%s.sindex' % SOURCE_TYPE_NAME
 
 Addr = namedtuple('Addr', ['host', 'port'])
+
+def parse_sindex_namespace(data, namespace):
+    idxs = []
+    while data != []:
+        l = data.pop(0)
+
+        match = re.match('^ns=%s:[^:]+:indexname=([^:]+):.*$' % namespace,l)
+        if match == None:
+            continue
+        idxs.append(match.groups()[0])
+
+    return idxs   
 
 class AerospikeCheck(AgentCheck):
 
@@ -41,6 +54,12 @@ class AerospikeCheck(AgentCheck):
                 for ns in namespaces:
                     conn.send('namespace/%s\r' % ns)
                     self._process_data(fp, NAMESPACE_EVENT_TYPE, namespace_metrics, tags+['namespace:%s' % ns])
+
+                    conn.send('sindex/%s\r' % ns)
+                    for idx in parse_sindex_namespace(fp.readline().split(';')[:-1], ns):
+                        conn.send('sindex/%s/%s\r' % (ns,idx))
+                        self._process_data(fp, SINDEX_EVENT_TYPE, [], 
+                                tags+['namespace:%s' % ns, 'sindex:%s' % idx])
 
                 conn.send('throughput:\r')
                 self._process_throughput(fp.readline().rstrip().split(';'), NAMESPACE_TPS_EVENT_TYPE, namespaces, tags)
--- a/aerospike/check.py
+++ b/aerospike/check.py
@@ -3,6 +3,8 @@
 
 # stdlib
 import socket
+import time
+import re
 from contextlib import closing
 from collections import namedtuple
 
@@ -13,6 +15,7 @@ SOURCE_TYPE_NAME = 'aerospike'
 SERVICE_CHECK_NAME = '%s.cluster_up' % SOURCE_TYPE_NAME
 CLUSTER_EVENT_TYPE = SOURCE_TYPE_NAME
 NAMESPACE_EVENT_TYPE = '%s.namespace' % SOURCE_TYPE_NAME
+NAMESPACE_TPS_EVENT_TYPE = '%s.namespace.tps' % SOURCE_TYPE_NAME
 
 Addr = namedtuple('Addr', ['host', 'port'])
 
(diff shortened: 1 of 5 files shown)


jeremy-lq:
@dmotylev has agreed to be the maintainer for this integration.

truthbk:
@dmotylev Thank you! I'll be reviewing this before the end of the week, so we get the ball rolling!

aerospike/test_aerospike.py
```diff
@@ -22,17 +22,25 @@
...
+                    'metrics':['cluster_size'],
+                    'namespace_metrics':['objects','hwm_breached'],
+                    'namespaces':{'test'},
+                    'tags':['tag:value']
+                }
+            ]
+        }
+        self.run_check(config)
+
+        self.assertMetric('aerospike.cluster_size', at_least=1)
```
truthbk: Can we assert that the tagging is also correct (so `['namespace:test', 'tag:value']`)?

aerospike/ci/aerospike.rake
```diff
@@ -8,21 +8,34 @@ def aerospike_rootdir
   "#{ENV['INTEGRATIONS_DIR']}/aerospike_#{aerospike_version}"
 end
 
+container_name = 'dd-test-aerospike'
```
truthbk: 🍰  thanks for setting these up 👌

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
+    def _send(self, event_type, key, val, tags={}):
+        datatype = 'event'
+
+        if re.match('^{(.+)}-^[-]+-hist',key):
+            return # skip histogram configuration
+
+        if key == 'cluster_key':
+            val = str(int(val, 16))
+
+        if val.isdigit():
```
truthbk: Is there any scenario where we might want to submit a `rate` instead of a `gauge`? I know that's not possible to detect in a generic way. But perhaps it's something we could add to the YAML as an attribute for a metric somehow (if you see a use for it). A `rate` would typically be useful for metrics that will always be increasing (for instance _number of requests since startup_, or whatever).

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
+        else:
+            required_data = d
+
+        for key, value in required_data.items():
+            self._send(event_type, key, value, tags)
+
+    def _send(self, event_type, key, val, tags={}):
+        datatype = 'event'
+
+        if re.match('^{(.+)}-^[-]+-hist',key):
```
truthbk: we should probably log something here... probably at the `debug` level would be enough if this is common, or `info` if it's uncommon.

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
-                    self.gauge(self._make_key(skey), value, tags=tags)
+            try:
+                float(val)
+                datatype = 'gauge'
+            except ValueError:
+                datatype = 'event'
+
+        if datatype == 'gauge':
+            self.gauge(self._make_key(event_type, key), val, tags=tags)
+        #else:
```
truthbk: We should either kill the `event` stuff or submit the events somehow....

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
+        d = dict(x.split('=', 1) for x in fp.readline().rstrip().split(';'))
+        if required_keys:
+            required_data = {k: d[k] for k in required_keys if k in d}
+        else:
+            required_data = d
+
+        for key, value in required_data.items():
+            self._send(event_type, key, value, tags)
+
+    def _send(self, event_type, key, val, tags={}):
```
truthbk: I think I would prefer some name other than `event_type`, particularly because I don't want anyone reading the code being misled into thinking `_send()` only deals with datadog event submission, etc. I think the same applies to `_process_data()` and `_process_throughput()`. Any chance we can rename this? :)

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
+        while data != []:
+            l = data.pop(0)
+
+            # skip errors
+            while l.startswith('error-'):
+                if data == []:
+                    return
+                l = data.pop(0)
+
+            match = re.match('^{(.+)}-([^:]+):',l)
```
truthbk: Could you please add in a comment above this line a sample line that would match? That would help understanding the remainder of the logic.

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
+from collections import namedtuple
 
 # project
 from checks import AgentCheck
 
-EVENT_TYPE = SOURCE_TYPE_NAME = 'aerospike'
-SERVICE_CHECK_NAME = '%s.server_up' % EVENT_TYPE
+SOURCE_TYPE_NAME = 'aerospike'
+SERVICE_CHECK_NAME = '%s.cluster_up' % SOURCE_TYPE_NAME
+CLUSTER_EVENT_TYPE = SOURCE_TYPE_NAME
```
truthbk: Are these really event types? Or are they more metric types?

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
+SINDEX_EVENT_TYPE = '%s.sindex' % SOURCE_TYPE_NAME
 
+Addr = namedtuple('Addr', ['host', 'port'])
+
+def parse_sindex_namespace(data, namespace):
+    idxs = []
+    while data != []:
+        l = data.pop(0)
+
+        match = re.match('^ns=%s:[^:]+:indexname=([^:]+):.*$' % namespace,l)
```
truthbk: Just for reference, could you post in a small comment a small sample match line for a matching namespace index line?

aerospike/check.py
```diff
@@ -3,75 +3,173 @@
...
+
+            key = match.groups()[1]
+            if data == []:
+                return # unexpected EOF
+            val = data.pop(0).split(',')[1]
+
+            self._send(event_type, key, val, tags + ['namespace:%s' % ns] )
+
+    def _process_data(self, fp, event_type, required_keys=[], tags={}):
+        d = dict(x.split('=', 1) for x in fp.readline().rstrip().split(';'))
```
truthbk: Just curious: are we guaranteed the data to be processed is always in a single line (ie. a single `readline()` will always suffice)?

aerospike/manifest.json
```diff
@@ -4,7 +4,8 @@
   "max_agent_version": "6.0.0",
```
truthbk: I believe you'd be happy to be the maintainer for the check 🙇  if that's still the case, could you update the `maintainer` field in the manifest with your email contact info? Thank you!

aerospike/manifest.json
```diff
@@ -4,7 +4,8 @@
...
   "min_agent_version": "5.6.3",
   "name": "aerospike",
-  "short_description": "aerospike description.",
+  "short_description": "Collect cluster and namespaces statistics",
+  "description": "The Aerospike integration helps to collect performance and avaialbility metrics from Aerospike cluster instances.",
   "support": "contrib",
   "supported_os": ["linux","mac_os","windows"],
   "version": "0.1.0"
```
truthbk: This is almost ready to go, so I would say we can set the version to `1.0.0` 🎉

aerospike/check.py
suthio: Yes! `asinfo` command is single line output

aerospike/check.py
suthio: yes. should be to event type.

aerospike/test_aerospike.py
suthio: I think that right.

jeremy-lq:
@dmotylev just want to touch base and see if you had a chance to look at the review comments left by Jaime.

dmotylev:
I'm going to check it later this week

jeremy-lq:
Thanks for the update, @dmotylev.

aerospike/check.py
whyrg: Quiet a few, we provide a mapping in our collectd plugin:
https://github.com/aerospike/aerospike-collectd/blob/develop/aerospike_schema.yaml

The "rate" type would be classified under "operations" in our schema file.

whyrg:
FYI, this PR is on our radar at Aerospike and we are keen to see it merged.

jeremy-lq:
Thanks, for touching base @whyrg. Once @dmotylev is able to address Jaime's review comments, we should be able to get this merged quickly.

truthbk:
@dmotylev seems like you got some people interested with your work! Any chance you could review the latest comments so we can make this readily available to the community? Thank you 🙇

jeremy-lq:
@whyrg has agreed to be the maintainer for this moving forward.

jeremy-lq:
Closing this PR in lieu of https://github.com/DataDog/integrations-extras/pull/110

PR #70, opened 2017-07-03
add nsd agent check

joewilliams:
### What does this PR do?

Adds a check for the nsd authoritative dns server daemon.

### Motivation

We use nsd.

### Testing Guidelines

### Versioning

- [ ] Updated `CHANGELOG.md`

### Additional Notes

### DIFF
--- a/nsd/check.py
+++ b/nsd/check.py
@@ -1,0 +1,40 @@
+# stdlib
+import os
+import re
+
+# project
+from checks import AgentCheck
+from config import _is_affirmative
+from util import Platform
+from utils.subprocess_output import get_subprocess_output
+
+class NsdCheck(AgentCheck):
+    # Stats info https://www.nlnetlabs.nl/projects/nsd/nsd-control.8.html
+
+    NSD_CTL = ['nsd-control', 'stats']
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances=instances)
+        if instances is not None and len(instances) > 1:
+            raise Exception("NSD check only supports one configured instance.")
+
+    def check(self, instance):
+        if instance is None:
+            instance = {}
+
+        host = instance.get('host')
+        tags = instance.get('tags', [])
+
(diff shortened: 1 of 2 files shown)


jeremy-lq:
@joewilliams has agreed to be the maintainer for this integration. Thanks for the contribution. We'll get it reviewed ASAP.

nsd/check.py
```diff
@@ -0,0 +1,40 @@
...
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances=instances)
+        if instances is not None and len(instances) > 1:
+            raise Exception("NSD check only supports one configured instance.")
+
+    def check(self, instance):
+        if instance is None:
+            instance = {}
+
+        host = instance.get('host')
+        tags = instance.get('tags', [])
```
hush-hush: We should add the host to the `tags` list. It would help splitting metrics by instances (ie previous comment).

Also `tags` should be int the `conf.yaml.example`. For example: https://github.com/DataDog/integrations-extras/blob/master/snmpwalk/conf.yaml.example#L18

nsd/check.py
```diff
@@ -0,0 +1,40 @@
...
+
+        nsd_out, _, _ = get_subprocess_output(self.NSD_CTL + ['-s', host], self.log)
```
hush-hush: `get_subprocess_output` might raise an `OSError` exception if `nsd-control` binary isn't found. You should catch it and log a clear error stating that the integration needs the binary.

nsd/check.py
```diff
@@ -0,0 +1,40 @@
...
+# project
+from checks import AgentCheck
+from config import _is_affirmative
+from util import Platform
+from utils.subprocess_output import get_subprocess_output
+
+class NsdCheck(AgentCheck):
+    # Stats info https://www.nlnetlabs.nl/projects/nsd/nsd-control.8.html
+
+    NSD_CTL = ['nsd-control', 'stats']
```
hush-hush: We should allow the customer to overwrite the path to `nsd-control` with an optional key in the configuration. This would be the default value.

jeremy-lq:
@joewilliams Just want to touch base on this. Have you had a chance to look at Maxime's feedback?

PR #86, opened 2017-11-20
Adding an intergration for Stardog Knowledge Graph

buzztroll:
### What does this PR do?

Adds a check for Stardog (http://www.stardog.com)

A brief description of the change being made with this pull request.

This check will call out to a stardog server and request status information from /admin/status.
A JSON document is returned and values are converted gauges that are consumed by DataDog

### Motivation

DataDog is a great way to monitor Stardog servers.

Customers requesting integration with DataDog

### Testing Guidelines

### Versioning

- [X ] Bumped the version check in `manifest.json`
- [ X] Updated `CHANGELOG.md`

### Additional Notes


jeremy-lq:
Thanks for the PR, @buzztroll! integrations-extras is a community-driven repository for integrations written and maintained outside Datadog. While support is on a best-effort basis, we're happy to help with code reviews and testing. The integrations won't be bundled with the agent, but will be installable on an individual basis.

buzztroll:
@jeremy-lq sounds good.  I would be happy to take on that responsibility.

stardog/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "maintainer": "support@stardog.com",
+  "manifest_version": "0.1.0",
+  "max_agent_version": "6.0.0",
+  "min_agent_version": "5.6.3",
+  "name": "stardog",
+  "short_description": "A Stardog data collector for DataDog.",
+  "guid": "1b32f0d4-49ef-40fb-aec3-365e4e7cd6ee",
+  "support": "contrib",
+  "supported_os": ["linux","mac_os"],
```
ISauve: Is this integrations supported on windows? If not, please note that in the `Compatibility` section of the README :)

stardog/metadata.csv
```diff
@@ -0,0 +1,52 @@
...
+stardog.dbms.mem.heap.max,gauge,,byte,,The maximum heap size,0,stardog,Max Heap Size
+stardog.dbms.mem.direct.buffer.used,gauge,,byte,,The amount of direct memory used,-1,stardog,Direct Mem Used
+stardog.dbms.mem.direct.buffer.max,gauge,,byte,,The maximum amount of direct memory that can be used,0,stardog,Direct Mem Max
+stardog.dbms.mem.direct.pool.max,gauge,,byte,,The maximum size of the direct memory pool,-1,stardog,Direct Pool Max
+stardog.dbms.mem.direct.pool.used,gauge,,byte,,The amount of memory used from the direct pool,-1,stardog,Direct Pool Used
+stardog.dbms.mem.direct.max,gauge,,byte,,The maximum amount of direct memory that can be used,0,stardog,Direct Mem Max
+stardog.dbms.mem.direct.used,gauge,,byte,,The amount of direct memory used,-1,stardog,Direct Mem Used
+stardog.databases.system.planCache.ratio,gauge,,percent,,Plan cache ration,0,stardog,Plan Cache Ratio
+stardog.databases.system.planCache.size,gauge,,bytes,,Plan cache size,0,stardog,Plan Cache Size
+databases.<database name>.txns.openTransactions,gauge,,unit,,The number of open transations on a given database,0,stardog,Open TX
```
ISauve: Two notes here:
1) This metric (and all those below) should also be prefixed by `stardog.`, otherwise it could be very hard to find these metrics
2)  We don't use variable metric names because they make dashboards really hard to build - the database name should be added as a tag instead, that way users can easily aggregate and compare data from across different databases. For example, the ideal use case here would be for someone to be able to simply look up `stardog.databases.txns.openTransactions` and then use tags to sort out data from the database(s) they want to look at

stardog/manifest.json
buzztroll: The integration should work on windows as well.  I'll fix that.

stardog/metadata.csv
buzztroll: Thanks @ISauve.  For some of the metrics listed we don't have a way to get them across the entire system and thus they are only available when scoped to a database.  Is there an elegant way to handle that situation?

ISauve: If I'm correct in understanding that you're referring to the metrics which are database-specific (`txns.openTransactions`, `.txns.speed`, `.queries.running`, `.queries.speed`, `.openConnections`) then what you've done here is perfect. This ideal solution is for the metric itself to be generalized so that it will appear if the data is present for _any_ database, and then the tag with the database name gives the specific scope of the data - which is exactly what you did!

PR #90, opened 2017-12-05
reboot_required check

montdidier:
### Description

Introduces a system check that will return a status depending on if a system requires a reboot after a software update. When exactly the status change is triggered, is configurable.

### Motivation

Many of my systems are configured to update software automatically but not to reboot automatically. The reboot is performed manually.

This occasionally means that updates requiring a system reboot are not applied in a timely manner. This little check will make some noise if a system fails to reboot within the defined time.


jeremy-lq:
Thanks for the PR, @montdidier!  integrations-extras is a community-driven repository for integrations written and maintained outside Datadog. While support is on a best-effort basis, we're happy to help with code reviews and testing. The integrations won't be bundled with the agent, but will be installable on an individual basis.

reboot_required/check.py
```diff
@@ -0,0 +1,61 @@
...
+        created_at_file = instance.get('created_at_file', self.CREATED_AT_FILE)
+        warning_days = int(instance.get('days_warning', 7))
+        critical_days = int(instance.get('days_critical', 14))
+
+        return self._get_status(critical_days, warning_days, self._days_since(reboot_signal_file, created_at_file))
+
+    def _days_since(self, reboot_signal_file, created_at_file):
+        if isfile(reboot_signal_file):
+            if isfile(created_at_file):
+                created_at = self._get_created_at(created_at_file)
```
hush-hush: Why create the `reboot-required.created_at` file and not directly stat `/var/run/reboot-required` ?

reboot_required/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+  "maintainer": "help@datadoghq.com",
```
hush-hush: If you accept to be the maintainer, this would need to be updated to one of your email address.

reboot_required/check.py
montdidier: reboot-required is frequently updated as packages are installed, as I understand it, stat will only ever return the last modified time on many systems (windows being the exception). Since what I really want to track is the creation time of this file since the last reboot I decided I would tackle tracking that point in time this way. There are other tricks one can use but this seemed the easiest.

reboot_required/manifest.json
montdidier: Sure no problem. I'll update this PR.

reboot_required/check.py
hush-hush: Make sense, I did not think about multiple updates.

montdidier:
OK. I've updated the maintainer. Let me know if this PR needs any more cleanups or changes.

hush-hush:
@montdidier I forgot one think during my last review: You need to add your integration to the travis matrix by editing `.travis.yml` (see: https://github.com/DataDog/integrations-extras/blob/master/.travis.yml#L38).

After that I'll merge your PR !

montdidier:
@hush-hush Hopefully done. I'm not 100% clear on what this does so please let me know if correct.

hush-hush:
@montdidier The travis matrix list tests to be run by Travis (your tests will now be run on each commit).

Thanks a lot for this contribution !

PR #93, opened 2017-12-27
Giladsh/filebeat check fix

giladsh1:
…e permissions issue

### What does this PR do?

- fixed registry file parsing error - the file contains a list of dictionaries, not a dictionary
- added and explicit file permission issue - since filebeat creates the file as root

### Motivation

the fails to run with filebeat 

### Testing Guidelines

### Versioning

- [ ] Updated `CHANGELOG.md`

### Additional Notes

### DIFF
--- a/filebeat/CHANGELOG.md
+++ b/filebeat/CHANGELOG.md
@@ -1,0 +1,7 @@
+# 0.1.1
+[giladsh]
+- fixed registrty file parsing error - the file contains a list of dictionaries, not a dictionary
+- added and explicit file permission issue - since filebeat creates the file as root
+
+# 0.1.0
+initial version
--- a/filebeat/check.py
+++ b/filebeat/check.py
@@ -6,7 +6,7 @@
 
 # 3rd party
 import os
-import simplejson
+import ast
 
 # project
 from checks import AgentCheck
@@ -27,15 +27,23 @@ class FilebeatCheck(AgentCheck):
 
         registry_contents = self._parse_registry_file(registry_file_path)
 
-        for item in registry_contents.itervalues():
+        for item in registry_contents:
+            self.log.debug("found item: %s" % item)
             self._process_registry_item(item)
 
     def _parse_registry_file(self, registry_file_path):
         try:
-            with open(registry_file_path) as registry_file:
-                return simplejson.load(registry_file)
-        except IOError:
-            self.log.warn('No filebeat registry file found at %s' % (registry_file_path, ))
+            # check that the collector user has permissions to read the file
+            if os.access(registry_file_path ,os.R_OK):
+                with open(registry_file_path) as registry_file:
+                    return ast.literal_eval(registry_file.read()) # convert the content from str to list
+            else:
+                # no permissions, alert and return empty dict
+                self.log.error("Cannot open the registry log file - check the file permissions")
+                return {}
+        except Exception as e:
+            self.log.error("General error while reading the registry file,"
+                           " the error message is: %s" % e.message)
             return {}
 
     def _process_registry_item(self, item):
@@ -57,3 +65,4 @@ class FilebeatCheck(AgentCheck):
 
     def _is_same_file(self, stats, file_state_os):
         return stats.st_dev == file_state_os['device'] and stats.st_ino == file_state_os['inode']
+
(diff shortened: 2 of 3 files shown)


hush-hush:
Hi @wk8,

Do you think you will have some time in the coming weeks to look at this ?

Thanks you !

filebeat/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# 0.1.1
+[giladsh]
+- fixed registrty file parsing error - the file contains a list of dictionaries, not a dictionary
```
wk8: Seems that changed with ELK v5.x, registry files were of the form https://github.com/DataDog/integrations-extras/pull/48/files#diff-9e185e1d6504fce30e41ef63f6bf4c18 before that.

So two things here:
* we probably want to maintain backward-compatibility with filebeat < 5.x format
* unit tests should test both formats

filebeat/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
...
+- added and explicit file permission issue - since filebeat creates the file as root
```
wk8: That's very unfortunate... and I can't seem to find anything in their doc as to how to change that :-/ any suggestions as to how to best do this in an automated way? In particular, does filebeat ever re-create the file after creating the first time?

filebeat/check.py
```diff
@@ -6,7 +6,7 @@
 
 # 3rd party
 import os
-import simplejson
+import ast
```
wk8: What was wrong with `simplejson` ? :)

filebeat/check.py
```diff
@@ -27,15 +27,23 @@ def check(self, instance):
...
             self._process_registry_item(item)
 
     def _parse_registry_file(self, registry_file_path):
         try:
-            with open(registry_file_path) as registry_file:
-                return simplejson.load(registry_file)
-        except IOError:
-            self.log.warn('No filebeat registry file found at %s' % (registry_file_path, ))
+            # check that the collector user has permissions to read the file
+            if os.access(registry_file_path ,os.R_OK):
```
wk8: space after comma, not before

wk8: And much more pythonic to ask for forgiveness, not permission, see below :)

filebeat/check.py
```diff
@@ -27,15 +27,23 @@ def check(self, instance):
...
+                with open(registry_file_path) as registry_file:
+                    return ast.literal_eval(registry_file.read()) # convert the content from str to list
+            else:
+                # no permissions, alert and return empty dict
+                self.log.error("Cannot open the registry log file - check the file permissions")
+                return {}
+        except Exception as e:
```
wk8: Bad form to catch any exception. Here the only expected type of exception is IOError, no point in catching more than that.

filebeat/CHANGELOG.md
wk8: (just checked, it seems it does re-create it periodically, so just changing the permissions in Chef or what not won't be enough, damn... how did you go around that?)

filebeat/check.py
wk8: And as to the permission thing, would be better to handle that here:
```
except IOError as e:
  if e.errno == errno.EACCES:
    self.log.error('Cannot open the registry log file - check the file permissions')
  else:
    ...
```

And on second thought since all you care about here is logging, there's no reason to do any kind of special handling for permission issues, just doing
```
except IOError as e:
  self.log.error('Cannot read the registry log file at %s: %s' % (registry_file_path, e))
```
will be plenty explicit enough:
```
Cannot read the registry log file at /var/lib/filebeat/registry: [Errno 13] Permission denied: '/var/lib/filebeat/registry'
```

filebeat/check.py
```diff
@@ -57,3 +65,4 @@ def _process_registry_item(self, item):
 
     def _is_same_file(self, stats, file_state_os):
         return stats.st_dev == file_state_os['device'] and stats.st_ino == file_state_os['inode']
+
```
wk8: ? :)

wk8:
@hush-hush @giladsh1 reviewed, thanks :)

jeremy-lq:
@giladsh1 Let us know if you have any questions about the code review. Once you address the issues raised we can get this PR merged.

wk8:
Should be superseded by https://github.com/DataDog/integrations-extras/pull/124

PR #94, opened 2017-12-29
UPS Battery Integration via upsc

platinummonkey:
### What does this PR do?

Adds UPS Device stats reporting via `upsc` client.

This assumes `nut`/`upsd` is configured correctly and running and only tested on linux systems.

### Motivation

Having UPS health and charge data is really helpful in a rack and in home.

### Testing Guidelines

### Versioning

- [x] Bumped the version check in `manifest.json`
- [x] Updated `CHANGELOG.md`

### Additional Notes

### DIFF
--- a/upsc/check.py
+++ b/upsc/check.py
@@ -29,7 +29,7 @@ class UpscCheck(AgentCheck):
         """
         try:
             results = subprocess.check_output(['upsc', '-l'], stderr=self.DEV_NULL)
-            return [r.lstrip().rstrip() for r in results.split('\n')]
+            return [r.strip() for r in results.split('\n')]
         except subprocess.CalledProcessError as e:
             self.log.error("Unable to query devices: %s", e)
             return []
@@ -47,7 +47,7 @@ class UpscCheck(AgentCheck):
             stats = {}
             for line in results.splitlines(False):
                 key, val = line.split(':', 1)
-                stats[key] = val.lstrip().rstrip()
+                stats[key] = val.strip()
             return stats
         except subprocess.CalledProcessError as e:
             self.log.error("Unable to query device %s" % name, e)
@@ -76,10 +76,10 @@ class UpscCheck(AgentCheck):
             if found_re:
                 continue
 
-            try:
-                value = float(v.lstrip().rstrip())
+            try:  # try a number conversion
+                value = float(v.strip())
                 results[k] = value
-            except Exception:
+            except Exception:  # this is a string value instead
                 if k == 'ups.status':
                     if v.lower().startswith('ol') or v.lower().startswith('on'):
                         results[k] = 1.0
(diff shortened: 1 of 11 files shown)


upsc/check.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def list_ups_devices(self):
+        """ Generate and return the list of configured devices.
+
+        :return: list of devices by name
+        :rtype: list[str]
+        """
+        try:
+            results = subprocess.check_output(['upsc', '-l'], stderr=self.DEV_NULL)
+            return [r.lstrip().rstrip() for r in results.split('\n')]
```
hush-hush: Why not just `r.strip()` ?

upsc/check.py
```diff
@@ -0,0 +1,137 @@
...
+        :type name: str
+        :return: raw results in simple form
+        :rtype: dict(str, str)
+        """
+        try:
+            results = subprocess.check_output(['upsc', name], stderr=self.DEV_NULL)
+            stats = {}
+            for line in results.splitlines(False):
+                key, val = line.split(':', 1)
+                stats[key] = val.lstrip().rstrip()
```
hush-hush: same here, we could use `strip`.

upsc/check.py
```diff
@@ -0,0 +1,137 @@
...
+            found_re = False
+            for r in self.excluded_re:
+                if r.match(k):
+                    found_re = True
+                    break
+            if found_re:
+                continue
+
+            try:
+                value = float(v.lstrip().rstrip())
```
hush-hush: same here, we could use `strip`.

upsc/check.py
```diff
@@ -0,0 +1,137 @@
...
+        """ Update Configuration tunables from instance configuration.
+
+        :param instance: Agent config instance.
+        :return: None
+        """
+        self.string_tags = list(self.DEFAULT_STRING_TAGS)
+        self.string_tags.extend(instance.get('string_tags', []))
+
+        self.additional_tags = []
+        self.additional_tags.extend(instance.get('tags', []))
```
hush-hush: You could replace those 2 lines by:

```python
self.additional_tags = instance.get('tags', [])
```

upsc/check.py
```diff
@@ -0,0 +1,137 @@
...
+
+        self.excluded = list(self.DEFAULT_EXCLUDED_TAGS)
+        self.excluded.extend(instance.get('excluded', []))
+        self.excluded_re = []
+        for excluded_regex in instance.get('excluded_re', []):
+            self.excluded_re.append(re.compile(excluded_regex))
+
+        self.excluded_devices = []
```
hush-hush: Same here

upsc/check.py
```diff
@@ -0,0 +1,137 @@
...
+                if r.match(k):
+                    found_re = True
+                    break
+            if found_re:
+                continue
+
+            try:
+                value = float(v.lstrip().rstrip())
+                results[k] = value
+            except Exception:
```
hush-hush: Could you add a comment explaining that we're handling string value here ?

PR #105, opened 2018-01-25
Add logz.io integration tile

dsahni:
### What does this PR do?

Adds the logz.io integration tile. Ported over to the new method of adding tiles, originally from this PR: https://github.com/DataDog/IntegrationTemplate/pull/22


logzio/.gitignore
```diff
@@ -0,0 +1 @@
+**/.DS_Store
```
gmmeyer: Can you fix the project's gitignore to ignore all DS_Store files instead of adding a new one here?

https://github.com/DataDog/integrations-extras/blob/master/.gitignore#L89

dsahni: Thanks for catching that. That `.gitignore` shouldn't be there

PR #106, opened 2018-01-25
Add Buddy integration tile

dsahni:
### What does this PR do?

Adds an integration tile for Buddy. Ported over to using the new method of adding tiles from the PR here: https://github.com/DataDog/IntegrationTemplate/pull/30


buddy/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "buddy",
+  "manifest_version": "1.0.0",
```
gmmeyer: Manifest Version is the version of the json. The version of the app should be a separate entry, `version`

dsahni: 👍 That field is indeed the version of the manifest json and not the app version, so `manifest_version` is appropriate

PR #110, opened 2018-01-31
Aerospike Integration

whyrg:
### What does this PR do?

Adds an integration check for Aerospike Database. Originally written by @joshk0 (DataDog/dd-agent#2219), then updated by @dmotylev (DataDog/integrations-extras#67)

### Motivation

Previous work is no longer maintained.

### Testing Guidelines

The plugin follows Testing Guide from DD.

### Versioning

Changes in manifest and/or changelog are not required, as previous work has not been finished.


jeremy-lq:
This is a continuation of https://github.com/DataDog/integrations-extras/pull/67 and @whyrg has agreed to be the maintainer. Can we get a review @truthbk?

truthbk:
@whyrg thank you so much for taking over. We really appreciate that. We're working hard to get the new agent out the door that has taken a lot of our focus, but will review ASAP. Thanks again!

whyrg:
Thanks for the update. Eagerly awaiting this to be accepted

whyrg:
How the review going? I've added `set` metrics so that forced a new CI run. Previous failed due to merging to the wrong branch.

jeremy-lq:
Agent 6 will be released as GA this week, so a review should be coming ASAP. Also note, community office hours are Wednesday if you have any specific questions. Thanks again.

https://github.com/DataDog/dd-agent/wiki/Community-Office-Hours

truthbk:
Finally reviewing this  @whyrg!!! Sorry for the delay!

aerospike/check.py
```diff
@@ -1,77 +1,205 @@
...
-        want_metrics = set(instance.get('metrics', []))
-        want_namespace_metrics = set(instance.get('namespace_metrics', []))
-        namespaces = instance.get('namespaces', None)
+            with closing(conn.makefile('r')) as fp:
+                conn.send('statistics\r')
+                self._process_data(fp, CLUSTER_METRIC_TYPE, metrics, tags=tags)
+
+                namespaces = self._get_namespaces(conn, fp, required_namespaces)
+
+                for ns in namespaces:
```
truthbk: How large can `namespaces` get? Trying to get an idea of the cardinality for the metric.... What could one expect `len(namespaces)` to be in a production cluster? Is it arbitrarily high? If so, we might have to set a soft cap.

aerospike/check.py
```diff
@@ -1,77 +1,205 @@
...
+                    conn.send('namespace/%s\r' % ns)
+                    self._process_data(fp, NAMESPACE_METRIC_TYPE, namespace_metrics, tags+['namespace:%s' % ns])
+
+                    conn.send('sindex/%s\r' % ns)
+                    for idx in parse_namespace(fp.readline().split(';')[:-1], ns,'indexname'):
+                        conn.send('sindex/%s/%s\r' % (ns,idx))
+                        self._process_data(fp, SINDEX_METRIC_TYPE, [], 
+                                tags+['namespace:%s' % ns, 'sindex:%s.%s' % (ns,idx)])
```
truthbk: if we're already sending `namespace:foo` does it make sense to use `foo` in `sindex` (ie. `sindex:foo.1`)? Or would sending `namespace:foo` and `sindex:1` suffice? Also, again to understand cardinality, what would be the expected order of magnitude for `sindex`?

aerospike/check.py
```diff
@@ -1,77 +1,205 @@
...
+                    
+                    conn.send('sets/%s\r' % ns)
+                    for s in parse_namespace(fp.readline().split(';'),ns,'set'):
+                        conn.send('sets/%s/%s\r' % (ns,s))
+                        self._process_data(fp, SET_METRIC_TYPE, [],
+                                tags+['namespace:%s' % ns, 'set:%s.%s' % (ns,s)], ':')
```
truthbk: ditto to the comment for `sindex` regarding the `set` tag. Also, again, what might be the expected order of magnitude for `sets`?

aerospike/test_aerospike.py
```diff
@@ -22,17 +22,25 @@
...
+                    'metrics':['cluster_size'],
+                    'namespace_metrics':['objects','hwm_breached'],
+                    'namespaces':{'test'},
+                    'tags':['tag:value']
+                }
+            ]
+        }
+        self.run_check(config)
+
+        self.assertMetric('aerospike.cluster_size', at_least=1)
```
truthbk: If we move the metric catalog to the check, it should be fairly straight-forward to iterate over some of the metrics to get some better metric catalog coverage? Would that be something you could entertain? I know I'm being greedy here, great work :)

aerospike/check.py
```diff
@@ -1,77 +1,205 @@
...
+
+        if re.match('^{(.+)}-(.*)hist-track',key):
+            self.log.debug("Histogram config skipped: %s=%s",key,str(val))
+            return # skip histogram configuration
+
+        if key == 'cluster_key':
+            val = str(int(val, 16))
+
+        if val.isdigit():
+            if key in self.init_config.get('mappings',[]):
```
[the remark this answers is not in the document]
whyrg: This is correct. I could include the entire mapping scheme, but that would balloon the size considerably.

aerospike/check.py
whyrg: https://www.aerospike.com/docs/guide/limitations.html

Specifically, Limit of 32 Namespaces (databases), 1023 sets (tables), and 32767 bins (columns).  We don't usually see more than 4 in production clusters.

aerospike/check.py
whyrg: Max of 256 sindexs per namespace. 

We can have the same sindex name across different namespaces.

You're asking about the tagging, right? Why have `namespace:foo` and `sindex:foo.bar` instead of `namespace:foo.bar`? I'm not entirely sure of DD's front end, but generally having 2 separate tags would mean I can have a partial match for metrics, instead of exact match.

aerospike/check.py
whyrg: 1023 sets per namespace

aerospike/conf.yaml.example
```diff
@@ -1,9 +1,357 @@
 init_config:
-  # Any global configurable parameters should be added here
+  # Rate mappings. DO NOT EDIT unless you know what you're doing.
+  # See https://github.com/aerospike/aerospike-collectd/blob/develop/aerospike_schema.yaml
+  # for the definitive list
+  mappings:
+    - basic_scans_failed    # moved to NS scan_basic_error
```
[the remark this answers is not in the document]
whyrg: I chose to have the mappings in the config file for easier revisions, and an option to the end-user, should they want it. Much easier and safer to change a config file than dig through code.

aerospike/check.py
```diff
@@ -1,77 +1,205 @@
...
+    def _get_namespaces(self, conn, fp, required_namespaces=[]):
+        conn.send('namespaces\r')
+        namespaces = fp.readline().rstrip().split(';')
+        if required_namespaces:
+            return [v for v in namespaces if v in required_namespaces]
+        else:
+            return namespaces
+
+    def _get_connection(self, addr):
+        conn = self.connections.get(addr, None)
```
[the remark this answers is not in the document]
whyrg: I guess the key issue is what does the parent context do when it receives an exception during `AgentCheck.check()`? Ideally, we'd retry `AgentCheck.check()`, which the current code assumes.

If the parent context doesn't like `check()` throwing an exception, I can just remove the `raise e` on line 95. Nothing gets sent, an exception is logged, and the agent waits for the next `check()` interval.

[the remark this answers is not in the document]
truthbk: You're 100% right - I somehow missed you were popping the connection after the exception! My bad! This is fine!

aerospike/check.py
truthbk: perfect, nothing to worry about here.

aerospike/check.py
truthbk: I actually mean why have `namespace:foo` `sindex:foo.bar` instead of `namespace:foo`, `sindex:bar`?

aerospike/check.py
truthbk: This is a bit high... I know this is the upper bound. Let me get back to you regarding this.

aerospike/conf.yaml.example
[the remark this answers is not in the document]
truthbk: OK, let's leave these all here. We're introducing a `metrics.yaml` configuration file that will come in handy with auto-discovery. We can open a PR when the time comes to possibly move those there.

aerospike/check.py
whyrg: Ah ok.

whyrg:
Still working on the CI tests and soft caps.

whyrg:
Oh my, the CI platform is on Windows.

whyrg:
@truthbk Let me know how'd you'd like to proceed. The tests use the Aerospike Python client, which is not compatible with Windows, but your AppVeyor CI is on Windows.

At this point my only option is to revert the test changes.

truthbk:
@whyrg let me work on fixing the CI - I have brought in a few changes from `master`. Don't revert your changes, this is really great stuff - I'll make sure we get it to work.

whyrg:
Your help is much appreciated!

truthbk:
@whyrg can you review the changes and see what you think? The test is now green on Appveyor, and running on Travis, so you should be able to continue working on it.

Also please notice I accidentally messed up the branch while pushing to your remote, so I had to fix it by force pushing. I'm really sorry about that. All the work is still there, and authorship is preserved. 

Let me know what you think about the state of the check, particularly the code I added to cap the number of `sindexs` and `sets`.

Thank you!

whyrg:
Perfect!

truthbk:
Aerospike is now passing the CI on Travis as well. I will get back to you later with any additional changes we might need to add (maybe none!)! Thanks for your patience @whyrg

aerospike/metadata.csv
```diff
@@ -0,0 +1 @@
+metric_name,metric_type,interval,unit_name,per_unit_name,description,orientation,integration,short_name
```
truthbk: @whyrg I think all we are missing is the metric metadata. It's probably the most boring part so no wonder we almost forgot about it! 😅 

You can look at some of the other integrations for guidance:
- metric_name (required): the name of the metric, plain and simple (eg. `aerospike.foo.bar`)
- metric_type (required): (note that as confusing as it sounds metrics we submit as rates should actually be listed as gauges in the CSV metric list, rates are a different kind of metric - TLDR; just set `gauge` for `rates` and `gauges`)  
   + gauge
   + counter
   + rate
   + count
- interval: (optional) The interval used for conversion between rates and counts. Since we don't have actual rates (remember what I mentioned above about rates being gauges?), you can probably leave this empty.
- unit_name (optional - but add if possible): the name of the units for the metrics.
- per_unit_name (optional): the name of the unit per type. For instance `blocks/second` would have `unit_name` `block`, and `per_unit_name`: `second`.
- description: (optional - but add if at all possible) A description of the metric. Limit of 400 characters. 
- orientation (required):
  + -1 - smaller values are better
  + 0 - no values are better or worse
  + 1 - larger values are better
- integration (required): this value will be `aerospike`
- short name (optional): a shortened version of the metric name.

If you can fill those in, I will then go over it and amend whatever requires fixing, and merge! 

Thanks for your help and patience @whyrg

truthbk: You can find some more details in the website: https://docs.datadoghq.com/developers/integrations/integration_sdk/#metadata-csv

whyrg: Sure thing. We have a lot of metrics so this will take a while.

whyrg:
metadata has been added. There's ~530 metrics, about 1/3 are deprecated duplicates to maintain backwards compatibility with older Aerospike Server versions.

PR #116, opened 2018-02-16
Add GNATSD integration

stephenprater:
This adds the GNATSD integration. @stephenprater and @jaredhoyt are
the maintainers of it.

This is being maintained under the auspices of Goldstar Events, Inc.

Co-authored-by: Jared Hoyt <jaredhoyt@gmail.com>
Co-authored-by: Stephen Prater <me@stephenprater.com>

### What does this PR do?

Add an integration for the NATS messaging server.

### Motivation

We use NATS at Goldstar - although NATS provides a monitoring endpoint, we would rather have this data at our fingertips in Datadog.

### Versioning

This is an initial version so doesn't seem necessary.

### Additional Notes

NATS Server is a simple, high performance open source messaging system for cloud native applications, IoT messaging, and microservices architectures.

We also plan on submitting an integration for NATS Streaming Server, a NATS "plugin" that provides durable message queues.


stephenprater:
I'm gonna quarrel with AppVeyor / Rubocop here.  Decomposing that rake task in order to make it two lines shorter is going to _hurt_ it's clarity of intent.

jeremy-lq:
Thanks for the PR, @stephenprater! integrations-extras is a community-driven repository for integrations written and maintained outside Datadog. While support is on a best-effort basis, we're happy to help with code reviews and testing. The integrations won't be bundled with the agent, but will be installable on an individual basis.

With that in mind, would you (and/or @jaredhoyt) be interested in being the initial maintainer for this integration? You'd be responsible for general code maintenance along with keeping an eye on PRs and issues. There's no specific time or duration commitment, but we do ask you only agree to become the maintainer if you feel it's something you'll be able to do for the foreseeable future. If you have any questions, let us know.

stephenprater:
@jeremy-lq Our team at Goldstar will maintain it, sure.  Do you need one of us to _individually_ agree to me be the maintainer in addition?

jeremy-lq:
@stephenprater, the team maintaining it works so no individual confirmation is necessary. We're working hard to get Agent 6 released as GA at the moment, but after that we'll get a code review so this can be merged ASAP. Thanks for the contribution and agreeing to have Goldstar maintain the integration.

stephenprater:
Everything is green here except the ApacheStorm travis flavor.

gnatsd/check.py
```diff
@@ -0,0 +1,150 @@
...
+
+class GnatsdCheck(AgentCheck):
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+        self.counts = {}
+
+    def check(self, instance):
+        self.InstanceCheck(instance, self).check()
+
+    class InstanceCheck:
```
gmmeyer: Can you merge this into the parent class? This class doesn't seem to add anything to the parent class other than another level of indirection.

gnatsd/check.py
```diff
@@ -0,0 +1,150 @@
...
+        TAGS = {
+            'varz': ['server_id'],
+            'connz.connections': ['cid','ip','name','lang','version'],
+            'routez.routes': ['rid','remote_id','ip']
+        }
+
+        def __init__(self, instance, checker):
+            self.instance = instance
+            self.checker = checker
+            self.config = self.Config(instance)
```
gmmeyer: `Config` seems like it could be replaced with a `dict`.

gnatsd/check.py
```diff
@@ -0,0 +1,150 @@
...
+
+        TAGS = {
+            'varz': ['server_id'],
+            'connz.connections': ['cid','ip','name','lang','version'],
+            'routez.routes': ['rid','remote_id','ip']
+        }
+
+        def __init__(self, instance, checker):
+            self.instance = instance
+            self.checker = checker
```
gmmeyer: Moving this up a level would remove the need to add this indirection.

gnatsd/check.py
```diff
@@ -0,0 +1,150 @@
...
+            self.config = self.Config(instance)
+            self.tags = self.config.tags + ['server_name:%s' % self.config.server_name]
+            self.service_check_tags = self.tags + ['url:%s' % self.config.host]
+
+        def check(self):
+            # Confirm monitor endpoint is available
+            self._status_check()
+
+            # Gather NATS metrics
+            for endpoint, metrics in self.METRICS.items():
```
gmmeyer: This is making a lot of api calls. Can you get this data in fewer calls? It's fine if you can't, no big deal. What we have to do we have to do. Just thought I'd ask :)

stephenprater: It should only be three API calls - each NATS monitoring endpoint includes different metrics - but as far as I am aware there's no _single_ API endpoint that returns them all.

gnatsd/check.py
stephenprater: Nah - it reads in the instance and then returns calculated values.  Otherwise we'd have to do those calculations in human heads.  If the values were constants I'd agree with you for sure.

gnatsd/check.py
stephenprater: The `GnatsdCheck` is a "singleton" that receives multiple calls to the `check` method with an _instance_ object.   Then we do multiple calls that check different things derived from that `instance` object.  This leaves us passing the `instance` argument around to every method and keeping track of the state in various method arguments.  It's really very complicated and procedural to maintain the state like that.   We found it much more straightforward to keep the "per call" state in an instance of an object (the `InstanceCheck`) and encapsulate the state of individual `check` calls outside of the GnatdsCheck singleton.

Is there a reason you'd prefer the more procedural style?

gnatsd/check.py
gmmeyer: That's fine! Just making sure!

gnatsd/check.py
gmmeyer: You could also do this:

```python
{
    'port': int(instance.get('port', 8222))
}
```

gnatsd/check.py
gmmeyer: Well, it's the way we do all of the other checks so it's a bit easier to follow. However, we've allowed people to use this style before. Instead of doing it like this, though, can you change a few things for me? First, can you give the class a better name. Second can you make it a top level class? The nested class is much harder to follow.

And it's the same with the config, just give it a better name and put it at the top level in the file. While I think using a dict for config would be a bit simpler, I don't really oppose this style of writing a check.

gnatsd/check.py
```diff
@@ -0,0 +1,153 @@
...
+        self.counts = {}
+
+    def check(self, instance):
+        self.InstanceCheck(instance, self).check()
+
+    class InstanceCheck:
+        SERVICE_CHECK_NAME = 'gnatsd.can_connect'
+
+        METRICS = {
+            'varz': {
```
wallyqs: List of metrics looks good 👍

gnatsd/ci/gnatsd.rake
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    task :install do
+      Rake::Task['ci:common:install'].invoke('gnatsd')
+      volumes = "--volume=#{gnatsd_temp_dir}/data:/usr/share/nats/data"
+      publish = '--publish=4222:4222 --publish=8222:8222'
+      cluster = '-cluster=nats://0.0.0.0:5222'
+
+      # Set up NATS cluster
+      sh %(docker run -d --name #{container_name}-serverA --publish=5222:5222 nats:1.0.4 #{cluster})
```
wallyqs: would not affect the test but `gnatsd` docker image `1.1.0` is already available if want to update to test against latest

stephenprater:
@gmmeyer All set!

PR #127, opened 2018-04-09
fix formatting and API key reference

richarddli:
Fix formatting and API key reference


dsahni:
lgtm @richarddli, will notify you when we have this available

PR #136, opened 2018-05-08
update logstash documentation

rsheyd:
Logstash check file is not included by default with the agent and thus requires extra configuration steps (downloading/renaming check) that are missing from docs.

### What does this PR do?

Adds required steps to download and rename logstash check file.

### Motivation

Users' questions about why logstash doesn't work after going through configuration steps.

### Testing Guidelines

### Versioning

- [ ] Updated `CHANGELOG.md`
- [x] If PR impacts documentation, docs team has been notified or [an issue has been opened on the documentation repository](https://github.com/DataDog/documentation/issues/new)

### DIFF
--- a/logstash/README.md
+++ b/logstash/README.md
@@ -11,6 +11,10 @@ Get metrics from Logstash service in real time to:
 
 ### Configuration
 
+Logstash is a community-supported integration so it's not packaged with the agent. 
+
+Download the [`check.py`][7] file, place it in the Agent's `checks.d` directory, and rename it to 'logstash.py'. 
+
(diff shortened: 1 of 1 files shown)


logstash/README.md
```diff
@@ -11,6 +11,10 @@ Get metrics from Logstash service in real time to:
 
 ### Configuration
 
+Logstash is a community-supported integration so it's not packaged with the agent.
```
jeremy-lq: We should come up with some standard wording here and put it in the main README, as it applies to all integrations-extras. Let's remove it from the PR and I'll add it there.

rsheyd: 👍

PR #138, opened 2018-05-12
[ci] fix docker pin

ofek:
### Motivation

Broken CI


ofek:
Solution: get rid of Ruby

masci:
@ofek please use `git commit --amend` and force push only one commit when playing with the CI. PRs are visible even when closed and this commit history is unacceptable.

PR #156, opened 2018-06-14
Revamp the CI

masci:
### What does this PR do?

Start moving the CI to use pytest

### Motivation

Consolidate the work we did in `integrations-core`


tasks/release.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # done
+    print("All done, remember to push to origin and open a PR to merge these changes on master")
+
+
+@task(help={
+    'target': "The check to release",
+    'dry-run': "Runs the task without publishing the package",
+})
+def release_upload(ctx, target, dry_run=False):
```
zippolyte: Should we keep this for extras since they are not packaged ?

tasks/utils/github.py
```diff
@@ -0,0 +1,55 @@
...
+# All rights reserved
+# Licensed under Simplified BSD License (see LICENSE)
+from __future__ import print_function, unicode_literals
+import os
+
+import requests
+
+
+API_URL = 'https://api.github.com'
+PR_ENDPOINT = API_URL + '/repos/DataDog/integrations-core/pulls/{}'
```
zippolyte: s/integrations-core/integrations-extras

tasks/utils/requirements.py
```diff
@@ -0,0 +1,56 @@
+# (C) Datadog, Inc. 2010-2017
```
zippolyte: Should we remove these requirement specific bits since it's not packaged in the agent ?

PR #168, opened 2018-06-20
Update check.py - Filebeat 6.3.0

danielkasen:
This is for 6.3 since they modified the layout

### What does this PR do?

Filebeat 6.3.0 now allows for registry file permissions, but the layout of the registry file is an array so the itervalues doesn't make sense and breaks on newer versions.

### Motivation
Getting this integration to work on filebeat > 6.0

### Additional Notes

This may break previous versions but I don't have sub 5.0 versions to test on.


phrawzty:
Hello,

Just to confirm, when you say "I don't have sub 5.0 versions to test on", does that mean you have tested it on 5.x? If yes, can you confirm that 5.x still works? Thanks!

danielkasen:
I don't have this version to test on at this moment.

ofek:
Included in https://github.com/DataDog/integrations-extras/pull/189

PR #171, opened 2018-07-02
[Eventstore] Added Integration

Xorima:
Feature: Adds the ability to capture eventstore metrics from datadog

### What does this PR do?

This PR introduces a new agent check for Eventstore, The open-source, functional database with Complex Event Processing in JavaScript.

### Motivation

Eventstore is a commonly used event processing database and the ability to see what is going on within any system is a high requirement

### Additional Notes

Not Applicable


jeremy-lq:
Thanks for the PR, @Xorima! integrations-extras is a community-driven repository for integrations written and maintained outside Datadog. While support is on a best-effort basis, we're happy to help with code reviews and testing. The integrations won't be bundled with the agent, but will be installable on an individual basis.

Xorima:
Happy to be a maintainer and keep an eye out @jeremy-lq

jeremy-lq:
Much appreciated. We'll have a code review from @DataDog/agent-integrations ASAP.

phrawzty:
Hello @Xorima! Thanks so much for this PR. All the bits and pieces are here, including tests, which is wonderful to see. We're just confirming the tests manually now and we'll get back to you asap.

cc @masci

eventstore/conf.yaml.example
```diff
@@ -0,0 +1,7 @@
+init_config:
```
masci: You can remove this file

eventstore/datadog_checks/eventstore/data/eventstore.yaml.default
```diff
@@ -0,0 +1,342 @@
+init_config:
```
masci: we're deprecating `init_config`, I'd move `default_timeout` and `tag_by_url` down at the `instance` level.
Regarding the `metric_definitions`, those should moved at the instance level as well, I'm assuming the list here contains the default metrics we always want to grab? In that case we could move their definition from the config file to the code.

eventstore/datadog_checks/eventstore/eventstore.py
```diff
@@ -0,0 +1,236 @@
...
+        default_timeout = self.init_config.get('default_timeout', 5)
+        timeout = float(instance.get('timeout', default_timeout))
+        tag_by_url = self.init_config.get('tag_by_url', False)
+        name_tag = instance.get('name', url)
+        metric_def = copy.deepcopy(self.init_config.get('metric_definitions', {}))
+        try:
+            r = requests.get(url, timeout=timeout)
+        except requests.exceptions.Timeout as e:
+            raise CheckException('URL: {0} timed out after {1} seconds.'.format(url, timeout))
+            return
```
masci: you can remove this `return`

eventstore/datadog_checks/eventstore/eventstore.py
```diff
@@ -0,0 +1,236 @@
...
+        except requests.exceptions.MissingSchema as e:
+            raise CheckException(e)
+        except requests.exceptions.ConnectionError as e:
+            raise CheckException(e)
+        # Bad HTTP code
+        if r.status_code != 200:
+            raise CheckException('Invalid Status Code, {0} returned a status of {1}.'.format(url, r.status_code))
+            return
```
masci: same

eventstore/datadog_checks/eventstore/eventstore.py
```diff
@@ -0,0 +1,236 @@
...
+        # Unable to deserialize the returned data
+        try:
+            parsed_api = json.loads(r.text)
+        except ValueError as e:
+            self.json_decode_event(url, r.text)
+            raise CheckException('{0} returned an unserializable payload of {1}.'.format(url, r.text))
+            return
```
masci: same

eventstore/setup.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Run-time dependencies
+    install_requires=get_requirements('requirements.in')+[
+        'datadog-checks-base',
+    ],
+    setup_requires=['pytest-runner', ],
+    tests_require=get_requirements('requirements-dev.txt'),
+
+    # Extra files to ship with the wheel package
+    package_data={'datadog_checks.eventstore': ['conf.yaml.example']},
```
masci: you can remove this line

eventstore/tests/test_check.py
```diff
@@ -0,0 +1,102 @@
...
+        c.check(instance)
+
+    # Decode Error
+    instance = {'url': 'https://google.com'}
+    with pytest.raises(CheckException):
+        c.check(instance)
+
+
+@pytest.fixture
+def aggregator():
```
masci: since this was already defined in `conftest.py`, pytest will make it automatically available to all the test modules, you can remove the function from here.

eventstore/tox.ini
```diff
@@ -0,0 +1,24 @@
+[tox]
```
masci: you never run tests marked with `integration`, is this intended?

masci:
@Xorima tests aren't doing anything at the moment because you should add `eventstore` here: https://github.com/DataDog/integrations-extras/blob/master/tasks/constants.py#L12 
Sorry for that, we're improving the test environment to make it more obvious for contributors (and for ourselves :) )

eventstore/datadog_checks/eventstore/data/eventstore.yaml.default
Xorima: I will look at doing this,
Metric Definitions are a global setting, it is giving the check knowledge of all the metrics, and means that if new metrics appear it would be easy to add them to the check without needing to fully understand the code.
I can add them to the code but it feels like it defeats the point here as I would like them globally accessible and easily changeable, do you have any other suggestions (Metrics seems too repetitive)

masci: We could use the same approach as the JMX-fetch based checks, let's say the `kafka` check.
The default set of metrics is kept in a special config file named `metrics.yaml`, see here: https://github.com/DataDog/integrations-core/blob/master/kafka/datadog_checks/kafka/data/metrics.yaml

Then, in the regular config file, one can specify that they want whatever metric is listed in `metrics.yaml` without any further configuration step: https://github.com/DataDog/integrations-core/blob/master/kafka/datadog_checks/kafka/data/conf.yaml.example#L62
or they could list explicitly what specific metric they want to collect.

The only difference would be keeping all of this at the `instance` level, for backward compatibility reasons we couldn't move metrics out of `init_config` for the kafka check.

Xorima:
Hi @masci 
I hope I managed to capture everything, I decided for the init_config to bring it into code, it will be a little bit more maintenance but it seems easier than adding a new dependency on pyyaml,
Thanks

Xorima:
This should be all ready for a CR now, thank you for waiting guys

PR #152, opened 2018-06-12
[logstash] Correct README.md link

lyletagawa:
### What does this PR do?

Corrects README documentation. Linked URI points to the wrong resource

### Motivation

The README documentation provides incorrect instructions

### Versioning

- [ ] Bumped the version check in `manifest.json`
- [ ] Updated `CHANGELOG.md`
- [ ] If PR impacts documentation, docs team has been notified or [an issue has been opened on the documentation repository](https://github.com/DataDog/documentation/issues/new)

### Additional Notes

PR #185, opened 2018-08-08
Fixes a bunch of links (see #152)

phrawzty:
### What does this PR do?

This is a clone of #152 with all conflicts resolved.


phrawzty:
Upon further review, it would appear that the changes were already made upstream - so this PR is now moot. Well, now we know. 😄

PR #194, opened 2018-08-17
Fixing filebeat integration for v 6.3.x

martonsz:
### What does this PR do?
[changelog/Fixed] Fixing filebeat integration so that it can read the new json structure for filebeat version 6.3.x
The registry file for filebeat has changed its structure and now uses an array of json objects instead of an array of keys
Tested with Filebeat 6.3.2

### Motivation
The current filebeat integration crashes when using Filebeat version >= 6.3.0


ofek:
Included in https://github.com/DataDog/integrations-extras/pull/189

PR #355, opened 2019-03-12
Support Python 3

ofek:
### Additional Notes

Also added e2e


traefik/setup.py
```diff
@@ -18,44 +18,36 @@
     long_description = f.read()
 
 
-# Parse requirements
-def get_requirements(fpath):
-    with open(path.join(HERE, fpath), encoding='utf-8') as f:
-        return f.readlines()
-
+CHECKS_BASE_REQ = 'datadog-checks-base'
```
zippolyte: Let's pin to `>=4.2.0` since we use the `base` submodule

PR #359, opened 2019-03-12
Support Python 3

therve:


sortdb/setup.py
```diff
@@ -18,11 +18,7 @@
     long_description = f.read()
 
 
-# Parse requirements
-def get_requirements(fpath):
-    with open(path.join(HERE, fpath), encoding='utf-8') as f:
-        return f.readlines()
-
+CHECKS_BASE_REQ = 'datadog-checks-base'
```
zippolyte: Let's pin this to `>=4.2.0` since we use `from datadog_checks.base` which is only valid for versions of the base check above 4.2.0

PR #385, opened 2019-04-03
Apply style changes

therve:


hithwen:
has this been done by a checkstyle script?

therve:
Yes we use black for that now. A new release of ddev happened, so now the integrations that have "dd_style_check" set need to have it applied.

PR #388, opened 2019-04-03
Fix integration tests

therve:
This passes the DOCKER env variables properly, and wait a bit more for the
environment to be up before testing.

### DIFF
--- a/riak_repl/tests/conftest.py
+++ b/riak_repl/tests/conftest.py
@@ -14,4 +14,4 @@ def dd_environment():
         conditions=[CheckEndpoints(URL, wait=5)],
         sleep=60
     ):
-        yield {'default_timeout': 5, 'url': URL}
+        yield
--- a/riak_repl/tests/conftest.py
+++ b/riak_repl/tests/conftest.py
@@ -10,9 +10,10 @@ from .common import HERE, URL
 @pytest.fixture(scope="session")
 def dd_environment():
     compose_file = os.path.join(HERE, "docker", "docker-compose.yml")
-    log_patterns = 'Full-sync with site "riak-west-1" completed'
+    log_patterns = ['Full-sync with site "riak-west-1" completed',
+                    'Full-sync with site "riak-east-1" completed']
     with docker_run(
         compose_file=compose_file,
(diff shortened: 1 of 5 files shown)


riak_repl/tests/conftest.py
```diff
@@ -1,15 +1,17 @@
...
 
 
 @pytest.fixture(scope="session")
-def riak_server():
+def dd_environment():
     with docker_run(
         compose_file=os.path.join(HERE, "docker", "docker-compose.yml"),
-        sleep=120
+        conditions=[CheckEndpoints(URL, wait=5)],
+        sleep=60
```
zippolyte: Is this sleep still needed then ?

riak_repl/tests/test_check.py
```diff
@@ -75,11 +74,7 @@ def test_service_check(aggregator, riak_server):
 
     c = RiakReplCheck('riak_repl', init_config, {}, None)
 
-    instance = {
-        'default_timeout': 5,
-        'url': 'http://{}:8098/riak-repl/stats'.format(get_docker_hostname())
-    }
-    c.check(instance)
+    c.check(dd_environment)
```
zippolyte: Can you make the configuration a constant and use it here instead of the return value of the fixture directly

riak_repl/tests/conftest.py
therve: Yes unfortunately it's still needed.

riak_repl/tests/test_check.py
therve: Can you explain a bit why? I thought one point of the fixture was to return the value that was going to be used for the check. Otherwise the return from the fixture is useless, no?

zippolyte: This specific fixture is a bit different. The returned value is used by `ddev` when spinning up the environment (the config yielded is then written to a file used as the configuration for the agent).

I wanted that mainly to do like we did in every other check. Also usually in the tests we use different instances to test for different things, that's why we don't use directly the return value of the fixture.

I guess for this one it works though. I don't have a strong opinion.

therve: That makes sense, let me change it.

therve:
Still flaky :/

therve:
Finally green!

PR #393, opened 2019-04-05
Add new metrics from stats endpoint

xkrt:
Add new metrics to Filebeat integration.

Looks like recent Filebeat versions changed some metric names. For example, Filebeat v6.7.0 do not expose in /stats metrics like:
`publish.events`
`libbeat.publisher.published_events`

But exposes metrics like:
`libbeat.pipeline.events.published`
`libbeat.output.events.acked`


phrawzty:
👋 Thanks for the PR @xkrt! We appreciate it!

PR #428, opened 2019-05-02
[Aqua] Migrating configuration file to new norm

l0k0ms:
### What does this PR do?

Migrates aqua configuration file to the new norm.


aqua/datadog_checks/aqua/data/conf.yaml.example
```diff
@@ -1,19 +1,27 @@
+init_config:
```
hkaj: Do we need init_config? I thought we wanted to make it optional with agent 6.

l0k0ms: I put it here for consistency.. idk what @DataDog/agent-integrations wants to do

nmuesch: All new integrations have it - https://github.com/DataDog/integrations-core/blob/master/datadog_checks_dev/datadog_checks/dev/tooling/templates/check/%7Bcheck_name%7D/datadog_checks/%7Bcheck_name%7D/data/conf.yaml.example#L1 but I think its highly dependent on whether the integration uses something from it.

hkaj: There's no need for it here, so if it's optional let's remove it.

FlorianVeaux: init_config is used in Agent 6 for the "common" configuration shared by all instances of the check. Some integrations like this one have no global configuration yet for consistency reasons `init_config` is kept (even empty) in most integrations.

Also a validator for those configuration files is on its way to be merged and integrated within the CI (see https://github.com/DataDog/integrations-core/pull/3679), which will enforce all conf files to have this `init_config` section.

hkaj: Works for me, didn't know we backtracked on the removal of init_config.

PR #447, opened 2019-05-03
[UPSC] Migrating configuration file to the new norm

l0k0ms:
### What does this PR do?

Migrating configuration file to the new norm.

### DIFF
--- a/upsc/datadog_checks/upsc/data/conf.yaml.example
+++ b/upsc/datadog_checks/upsc/data/conf.yaml.example
@@ -1,16 +1,47 @@
 init_config:
 
 instances:
+
+    ## @param ups - string - required
+    ##
+    #
   - ups: all
-  #   # your own custom tags
-  #   tags: ['custom:tag']
-  #   # string-based stats results that should be converted to tags on the metrics
-  #   string_tags: ['ups.beeper.status']
-  #   # specific list of stats to ignore
-  #   excluded: ['ups.load']
-  #   # regex list of stats to ignore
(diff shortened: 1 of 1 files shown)


upsc/datadog_checks/upsc/data/conf.yaml.example
```diff
@@ -1,16 +1,47 @@
 init_config:
 
 instances:
+
+    ## @param ups - string - required
```
l0k0ms: This parameter seems never used by the check.. should we remove it?

platinummonkey: it was a more of a way to give YAML something. We could change this to `alias` and allow the users to override the device name, which is sometimes ugly looking from the host.

upsc/datadog_checks/upsc/data/conf.yaml.example
```diff
@@ -1,16 +1,47 @@
...
-  #   # specific list of stats to ignore
-  #   excluded: ['ups.load']
-  #   # regex list of stats to ignore
-  #   excluded_re: ['ups.*']
-  #   # list of specific excluded devices
-  #   excluded_devices: ['badups1']
-  #   # regex list of excluded devices
-  #   excluded_devices_regex: ['ext.*']
+
+    ## @param tags - list of key:value element - optional
```
FlorianVeaux:
```suggestion
    ## @param tags - list of key:value elements - optional
```

PR #455, opened 2019-05-20
Windows containers integration

jaycdave88:
Windows Docker.

### What does this PR do?

This PR creates the ability to monitor Windows Containers to collect Infrastructure based metrics, which are reported by Windows Containers. 

### Motivation

My love for Windows knows no limits :) 

### Review checklist

- [x] Git history is clean
- [x] If PR impacts documentation, docs team has been notified or an issue has been opened on the [documentation repo](https://github.com/DataDog/documentation/issues/new)

### Additional Notes

I am an on the Sales Engineering team at Datadog.


windows_docker/src/win_docker.py
```diff
@@ -0,0 +1,120 @@
...
+		containers_response = requests.get(container_url, timeout=1).json()
+
+
+		for value in containers_response:
+			container_full_ids.append(value["Id"])
+			container_count = len(containers_response)
+			
+			if value:
+				id_tag = "sources:docker_win"
+				image_tag = "docker_image:" + value["Image"]
```
hkaj: is the Image field in this json always fully resolved? Sometimes in docker inspect we see things like `"datadog/agent@sha256:5fbe2c87d4c4c2d9fc2d4bf45e39b0572f7f3f8ff906e3a77515a3308ab0c2cf"`, in which case we need to resolve the image tags, see https://github.com/DataDog/dd-agent/blob/master/utils/dockerutil.py#L571-L615

hkaj: Also could you follow the convention? docker_image is the full image name, image_name is the name without the tag, image_tag is the opposite, and short_image is the image_name without the registry, if there is one. See https://github.com/DataDog/datadog-agent/blob/master/pkg/util/containers/image.go#L13-L45

windows_docker/src/win_docker.py
```diff
@@ -0,0 +1,120 @@
...
+				state_tag = "state:" + value["State"]
+				status_tag = "status:" + value["Status"]
```
hkaj: What do the values of state and status look like? It might be worth calling `.lower()` here.

windows_docker/src/win_docker.py
```diff
@@ -0,0 +1,120 @@
...
+				name_tag = "container_name:" + value["Names"][0]
+				self.gauge("docker.windows.created", value["Created"], tags=[id_tag, image_tag, name_tag, status_tag, state_tag ])
```
hkaj: Is this a metric that reports a timestamp? These are not super actionable. It might be better to throw an event for creating events?

windows_docker/src/win_docker.py
```diff
@@ -0,0 +1,120 @@
...
+	# running containers:
+		self.gauge("docker.windows.containers.running", container_count, tags=[id_tag, image_tag, name_tag, status_tag, state_tag])
```
hkaj: Generally speaking it would be useful to retrieve the same metrics (and keep the same name) as on Linux, so that users who run hybrid environments can display values of both linux and windows side by side, splitting them only with tags.

windows_docker/src/win_docker.py
```diff
@@ -0,0 +1,120 @@
...
+
+		for x in container_full_ids:
+			container_urls.append("http://localhost:2375/containers/" + x + "/stats")
```
hkaj: This shouldn't be hardcoded here. Let's use the config to get the connection info, and hardcode the endpoint part as a module constant.

hkaj: Also, what does this path look like, performance wise? I think this endpoint is basically the `inspect` command, and it can be quite slow, and expensive to serve for the docker daemon. How long does this part take if you run a few hundred containers? Also interested in the CPU/memory profile of the docker daemon when it serves these requests.

windows_docker/src/win_docker.py
jaycdave88: Gotcha, okay - I can remove all the `.windows.` from the metrics and just add a tag for os: windows, does that sound okay - or is there a specific tag that you would like me to reference?

windows_docker/src/win_docker.py
jaycdave88: Looking for more guidance around what a Docker event would look like - I have it stubbed out within my second code change. If you can take a look and let me know that would be great.

windows_docker/src/win_docker.py
jaycdave88: Still looking into how to achieve this.

windows_docker/datadog_checks/windows_docker/data/conf.yaml.example
```diff
@@ -0,0 +1,126 @@
+init_config:
```
l0k0ms: Hello :) In order for the CI tests to pass, you must follow the configuration file norm described here: https://docs.datadoghq.com/developers/integrations/new_check_howto/#configuration-file

l0k0ms: hello :) 

So Every commented line should at least have two `#` in front of it.
Each parameter should be preceded with a  `@param <name> - <type> - optional - default: <defval>` line

You may find an example of the expected configuration file shape here: https://docs.datadoghq.com/developers/integrations/new_check_howto/#example-configuration 

Feel free to get back to me if you have any questions :)

dabcoder: `7875b2a` attempts to fix the config file formatting.

PR #482, opened 2019-09-20
Add codeowner to riak

hithwen:
Hallo @abtreece, as the creator of the integration you may wanna become owner of it


abtreece:
I like it. 😁

PR #483, opened 2019-09-20
[ci] Use templates from core

FlorianVeaux:


FlorianVeaux:
Don't worry about Storm, just looks like github is broken 🤷‍♂ 
This file should be a 160MB file, not a small text file: https://github.com/platinummonkey/static_ci_files/blob/master/datadog/integrations-extras/apache-storm-1.2.1.tar.gz

PR #484, opened 2019-09-24
Updating instructions to match K8s v1.16

l0k0ms:
### What does this PR do?

Updates the instructions in order to match the K8s v1.16 new logic.


FlorianVeaux:
You can ignore the Storm test as it is failing for an external reason right now :)

PR #487, opened 2019-09-27
Fix storm link for tests

FlorianVeaux:
Storm link was hosted on github LFS. We reached the quota and can't download it anymore.

Let's use one of apache mirrors directly

### DIFF
--- a/storm/tests/compose/Dockerfile
+++ b/storm/tests/compose/Dockerfile
@@ -1,9 +1,9 @@
 FROM maven:3.5.0-jdk-7
 
 
-RUN git clone git://github.com/apache/storm.git \
-    && cd storm/examples/storm-starter \
-    && git checkout v1.2.3 \
+RUN curl -fsSL https://github.com/apache/storm/archive/v1.2.3.tar.gz -o storm.tar.gz \
+    && tar -xf storm.tar.gz \
+    && cd storm-*/examples/storm-starter \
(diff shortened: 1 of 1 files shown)


FlorianVeaux:
I just opened the PR to run the tests on the pipeline, but yes likely better to mirror it on S3

FlorianVeaux:
Waiting on https://github.com/DataDog/integrations-core/pull/4657 otherwise tests can't run correctly

FlorianVeaux:
Tests are good, ready for review. I didn't want to have yet another external dependency. Assuming the github link shouldn't change as fast as the apache mirror

storm/tests/compose/Dockerfile
```diff
@@ -1,10 +1,10 @@
 FROM maven:3.5.0-jdk-7
 
-RUN curl -fsSL https://github.com/platinummonkey/static_ci_files/raw/master/datadog/integrations-extras/apache-storm-1.2.1.tar.gz -o storm.tar.gz \
```
AlexandreYang: Question.

Any reason not using the release archives (https://github.com/apache/storm/releases) ?  e.g. `https://github.com/apache/storm/archive/v1.2.3.tar.gz`

Probably lighter to download than the git repository (and maybe more stable too).

platinummonkey: the mirror links are not reliable, they've been dropped before $0.02

PR #519, opened 2020-01-13
Remove image license

FlorianVeaux:
There is now a dedicated icon for this integration

PR #520, opened 2020-01-13
Fix storm metadata naming

FlorianVeaux:
Storm metadata are neither valid nor useful (as they need to match exactly what the backend has).
Replacing all `<interval>` in the name by `60` (the default interval).

People using custom interval values won't have the correct metadata, but one cannot add an entry per interval so 🤷‍♂ At least we should be able to sync metadata now.


florimondmanca:
Isn't this caught by any `ddev validate` command?

FlorianVeaux:
It is now, since https://github.com/DataDog/integrations-core/pull/5437
This was actually caught when doing this: https://github.com/DataDog/integrations-extras/pull/519

platinummonkey:
there are only a few intervals, should probably add them all honestly. 60, 30 and 5 if I recall correctly

PR #529, opened 2020-01-28
Fix check structure

AlexandreYang:
### What does this PR do?

Looks like `contrastsecurity` is only needed for logs and assets.

Hence, it does not need code related to python integration.

cc @kristianamitchellcontrastsecurity

### Review checklist

- [ ] Git history is clean

### Additional Notes


florimondmanca:
Sounds legit to me, since this integration only seems to provide instructions for setting up logs.

@AlexandreYang We need to make sure removing these files still allow the Agent to detect the `contrastsecurity` integration, i.e. reading the `contrastsecurity.d/conf.yaml` file.

AlexandreYang:
> We need to make sure removing these files still allow the Agent to detect the contrastsecurity integration, i.e. reading the contrastsecurity.d/conf.yaml file.

Just `contrastsecurity.d/conf.yaml` file alone will allow agent to use the conf file for logs. The name of the `integration` will be `contrastsecurity` (extracted from `contrastsecurity.d`)

contrastsecurity/datadog_checks/contrastsecurity/conf.yaml.example
```diff
@@ -1,20 +0,0 @@
-init_config:
-
-instances:
-
-## Log Section
```
ChristineTChen: > Just contrastsecurity.d/conf.yaml file alone will allow agent to use the conf file for logs

@AlexandreYang we don't need to keep this default example?

AlexandreYang: The issue is that, for `extras`, for this file to be copied, we need to created a wheel and then use `agent integration install` command. Which is not possible since there is no python integration anymore.

PR #532, opened 2020-02-05
Update .coveragerc

ofek:
### Motivation

https://github.com/DataDog/integrations-core/pull/5647

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,7 @@
 [run]
 branch = True
+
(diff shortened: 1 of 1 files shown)


.coveragerc
```diff
@@ -1,5 +1,6 @@
 [run]
 branch = True
+parallel = True
```
AlexandreYang: Could you add a comment that this has been added as workaround and link to the original issue?

PR #577, opened 2020-04-08
Fix metric_to_check for all integrations

FlorianVeaux:
We've enforced validation of the `metric_to_check` field of the manifest.json to be a metric that is documented as part of the metadata.csv.

A few integrations were either not specifying a correct metric or were specifying a metric when the integration does not submit any.

One can confirm that validation works by running `ddev -e validate manifest`.
The CI will still fail at the moment for another issue caused by the `pihole` config file being out of sync.


sigsci/manifest.json
```diff
@@ -6,8 +6,8 @@
   "maintainer": "info@signalsciences.com",
   "display_name": "Signal Sciences",
   "short_description": "Collect data from Signal Sciences to see anomalies and block attacks",
-  "metric_prefix": "sigsci.",
```
florimondmanca, line 9: What's the rationale for removing this option? (The value seems correct both here and in `rbltracker`)

FlorianVeaux: For `rbltracker`, the integration doesn't submit any metric at all.
For `sigsci` this is an oversight, I assumed the same because of the empty `metadata.csv`

FlorianVeaux: Updated

florimondmanca:
CI still failing due to #576 - let's try to get that in first.

Issue #612, opened 2020-06-08
[nvml] broken on p2 (K80) instances


therc:
The check is making calls without attempting to catch exceptions. This fails on old K80 machines such as p2 instances, because some metrics are not supported:

```
2020-06-08 17:32:11 UTC | CORE | ERROR | (pkg/collector/runner/runner.go:292 in work) | Error running check nvml: [{"message": "Not Supported", "traceback": "Traceback (most recent call last):
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/base/checks/base.py\", line 820, in run
    self.check(instance)
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/nvml/nvml.py\", line 81, in check
    self.gather(instance)
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/nvml/nvml.py\", line 94, in gather
    self.gather_gpu(handle, tags)
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/nvml/nvml.py\", line 122, in gather_gpu
    consumption = NvmlCheck.N.nvmlDeviceGetTotalEnergyConsumption(handle)
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/pynvml/nvml.py\", line 1252, in nvmlDeviceGetTotalEnergyConsumption
    check_return(ret)
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/pynvml/nvml.py\", line 366, in check_return
    raise NVMLError(ret)
pynvml.nvml.NVMLError_NotSupported: Not Supported
"}]
```

cep21:
Hi, are there other log messages?  Do you see "Unable to execute NVML function"?  What are the lines before and after?

I would expect that exception to be caught by https://github.com/DataDog/integrations-extras/blob/master/nvml/datadog_checks/nvml/nvml.py#L50

therc:
Unfortunately, that's all it spits out, at least according to Datadog logs, every 45s. A hacky workaround for now is to comment out lines 120-140.

cep21:
If you add prints inside `NvmlCall.__exit__` do you see which path the code is taking, or which `return` is getting called?

therc:
Is the exception type pynvml.nvml.NVMLError, perhaps?
I can't easily add debug statements right now, but I'll give it a shot and report here.

[...]

PR #613, opened 2020-06-08
Fix NVMLError subclass check

therc:
### What does this PR do?

It should catch NVMLError exceptions correctly.
Because the current code only performs an equality check, any subclass of NVMLError will never match, so the code will always return False.

### Motivation

As reported in #612, the check fails with an exception on older hardware.

### Review checklist

- [ ] Git history is clean

### DIFF
--- a/nvml/datadog_checks/nvml/nvml.py
+++ b/nvml/datadog_checks/nvml/nvml.py
@@ -43,7 +43,7 @@ class NvmlCall(object):
 
     def __init__(self, name, logger):
         self.name = name
-        self.log  = logger
+        self.log = logger
 
     def __enter__(self):
         pass
--- a/nvml/datadog_checks/nvml/nvml.py
+++ b/nvml/datadog_checks/nvml/nvml.py
@@ -51,7 +51,7 @@ class NvmlCall(object):
         # Do nothing if the exception is not from pynvml or there is no exception
         if traceback is None:
             return False
-        if exception_type is not pynvml.NVMLError:
+        if not issubclass(exception_type, pynvml.NVMLError):
             return False
 
         # Suppress pynvml exceptions so we can continue
--- a/nvml/datadog_checks/nvml/nvml.py
+++ b/nvml/datadog_checks/nvml/nvml.py
@@ -14,6 +14,7 @@ from datadog_checks.base import AgentCheck
 from .api_pb2 import ListPodResourcesRequest
 from .api_pb2_grpc import PodResourcesListerStub
 
+previously_printed_errors = set()
 METRIC_PREFIX = "nvml."
 SOCKET_PATH = "/var/lib/kubelet/pod-resources/kubelet.sock"
 """Assumed to be a UDS accessible from this running code"""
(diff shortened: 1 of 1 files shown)


nvml/datadog_checks/nvml/nvml.py
```diff
@@ -51,7 +51,7 @@ def __exit__(self, exception_type, exception_value, traceback):
         # Do nothing if the exception is not from pynvml or there is no exception
         if traceback is None:
             return False
-        if exception_type is not pynvml.NVMLError:
+        if not issubclass(exception_type, pynvml.NVMLError):
```
ofek, line 54:
```suggestion
        if not isinstance(exception_type, pynvml.NVMLError):
```

therc: Thanks, I also found another bug...

therc:
That did the trick and it only spews out two errors now, but they're ugly for unrelated reasons:

```
Error running check nvml: [{"message": "'NvmlCall' object has no attribute 'log'",
 "traceback": "Traceback (most recent call last):
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/nvml/nvml.py\", line 122, in gather_gpu
    consumption = NvmlCheck.N.nvmlDeviceGetTotalEnergyConsumption(handle)
Error running check nvml: [{"message": "'NvmlCall' object has no attribute 'log'",
 "traceback": "Traceback (most recent call last):
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/nvml/nvml.py\", line 137, in gather_gpu
    tx_bytes = NvmlCheck.N.nvmlDeviceGetPcieThroughput(handle, pynvml.NVML_PCIE_UTIL_TX_BYTES)
```

NvmlCall inherits from object and thus has no log field. Should it inherit from AgentCheck or something else instead?

ofek:
Do https://github.com/DataDog/integrations-extras/blob/b470681af0579da9fad20a1f3a351621e3861f14/nvml/datadog_checks/nvml/nvml.py#L44-L45

```

    def __init__(self, name, log):
        self.name = name
        self.log = log
```

therc:
I also updated all the calling sites. Unless you're saying there's some other special magic.

therc:
> I also updated all the calling sites. Unless you're saying there's some other special magic.

Now it only logs the first of the two errors, but it does so over and over....

therc:
I flipped it back to issubclass() and it runs as intended (log edited for clarity):
```
2020-06-09 01:08:21 | INFO | Running check nvml
2020-06-09 01:08:21 | INFO | Monitoring kubelet tags at /var/lib/kubelet/pod-resources/kubelet.sock
2020-06-09 01:08:21 | WARN | Unable to execute NVML function: total_energy_consumption: Not Supported
2020-06-09 01:08:21 | WARN | Unable to execute NVML function: pci_through: Not Supported
2020-06-09 01:08:21 | INFO | Done running check nvml
2020-06-09 01:08:36 | INFO | Running check nvml
2020-06-09 01:08:36 | INFO | Done running check nvml
2020-06-09 01:08:51 | INFO | Running check nvml
2020-06-09 01:08:51 | INFO | Done running check nvml
```

PR #624, opened 2020-06-19
Add support for redis_sentinels with auth requirements

Silthias:
### What does this PR do?

Adds a new config value to the redis_sentinel check of 'sentinel_password', if defined then the redis connection will use it, and if not then it sets to None.

Alters the config setup declaration to allow for unit testing and validation of expected behaviour.

Bumped relevant version files to 1.1.0

### Motivation

Requirement to monitor a sentinel setup with authentication, current implemention does not allow that.

### Review checklist

- [ ] Git history is clean

### Additional Notes


redis_sentinel/datadog_checks/redis_sentinel/redis_sentinel.py
```diff
@@ -13,9 +13,33 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
 
+    def _load_config(self, instance):
+        host = instance.get('sentinel_host')
+        port = instance.get('sentinel_port')
+
+        if not host or not port:
+            raise ConfigurationError('Configuration error, please fix redis_sentinel configuration')
+
+        try:
+            if float(port) != int(port):
```
ChristineTChen, line 24:
```suggestion
            if isinstance(port, int):
```

Let's directly check if the port is an int so we don't need the try/except block with duplicate error message.

redis_sentinel/datadog_checks/redis_sentinel/redis_sentinel.py
```diff
@@ -13,9 +13,33 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+                raise ConfigurationError('Configuration error. "sentinel_port" must be an int.')
+        except Exception:
+            raise ConfigurationError('Configuration error. "sentinel_port" must be an int.')
+
+        if instance.get('sentinel_password'):
+            passwd = instance.get('sentinel_password')
+        else:
+            passwd = None
```
ChristineTChen, line 32:
```suggestion
        passwd = instance.get('sentinel_password', None)
```

Silthias: Valid, thanks for this and implemented.

redis_sentinel/datadog_checks/redis_sentinel/redis_sentinel.py
Silthias: Valid, thanks for this and implemented.

PR #693, opened 2020-08-11
Sync config specs

AlexandreYang:
Sync after changes to base configs `instances/http`


AlexandreYang:
@mgarabed yes, will be fixed by https://github.com/DataDog/integrations-extras/pull/694

PR #736, opened 2020-10-14
Add config spec

AlexandreYang:
Use config spec.

This PR only change wording and add missing attributes to `conf.yaml.example`

cc @brentm5

### DIFF
--- a/resin/assets/configuration/spec.yaml
+++ b/resin/assets/configuration/spec.yaml
@@ -1,0 +1,19 @@
+name: Resin
+files:
+- name: resin.yaml
+  options:
+  - template: init_config
+    options:
+      - template: init_config/jmx
+  - template: instances
+    options:
+    - template: instances/jmx
+      overrides:
+        host.value.example: localhost
+        port.value.example: 7199
+  - template: logs
+    example:
+    - type: file
+      path: /var/opt/resin/log/*.log
+      source: resin
+      service: <SERVICE_NAME>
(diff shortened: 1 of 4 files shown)


florimondmanca (approved):
:+1:

Let's also double-check the base requirement is sufficient in `setup.py`, must be at least `datadog-checks-base>=11.0.0`.

florimondmanca (changes requested):
Update: looks like the base requirement should indeed be updated

https://github.com/DataDog/integrations-extras/blob/06d4039d299e41db776ab60de8afcb95c34f3776/resin/setup.py#L27

AlexandreYang:
@florimondmanca What change makes the change not compatible with older base dep ?

AlexandreYang:
@florimondmanca thx, updated the base req https://github.com/DataDog/integrations-extras/pull/736/commits/81f388856fbdf3ed8bb46f54996acb36d26fbf64

resin/assets/configuration/spec.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - template: instances/jmx
+      overrides:
+        host.value.example: localhost
+        port.value.example: 7199
+  - template: logs
+    example:
+    - type: file
+      path: /var/opt/resin/log/*.log
+      source: resin
+      service: <SERVICE_NAME>
```
ChristineTChen, line 19:
```suggestion
```

ChristineTChen (changes requested):
nit

ChristineTChen (approved):
thanks! LGTM

PR #738, opened 2020-10-14
Use real upstream repo in nvml test Dockerfile

therc:
Anyone trying to use the Dockerfile under nvml/tests to build a custom image with the integration built-in, will be building the container against a 3rd party repo, not Datadog's. Fix that.


ChristineTChen (approved):

ChristineTChen:
CI is failing for separate issue that has a pending fix https://github.com/DataDog/integrations-extras/pull/736

PR #769, opened 2020-12-09
Revert n2ws dash ids update

FlorianVeaux:
Some dashboards ids were changed in https://github.com/DataDog/integrations-extras/pull/760.

We can’t change the dashboard “ids” (defined in manifest.json) but we can change the dashboard names (defined in each dashboard payload).
This is a bit of a technical limitation of our backend, we can’t update identifiers without manually modifying the Postgres database manually
4:19
This does not have any visible impact, the ids is used as a reference and is never displayed in the UI


FlorianVeaux:
Renaming happened in https://github.com/DataDog/integrations-extras/pull/760

hithwen (approved):

PR #811, opened 2021-02-24
Pin docker image

hithwen:
Tests are failing because the docker image has changed

### DIFF
--- a/nextcloud/tests/compose/nextcloud-standalone
+++ b/nextcloud/tests/compose/nextcloud-standalone
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   nextcloud-standalone:
-    image: "nextcloud:latest"
+    image: "nextcloud:14.0.14"
     ports:
       - 8080:80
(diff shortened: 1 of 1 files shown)


nextcloud/tests/compose/nextcloud-standalone
```diff
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   nextcloud-standalone:
-    image: "nextcloud:latest"
+    image: "nextcloud:14.0.14"
```
eplanet, line 5: That's an image from Sept 2019. I guess that's fine for now to fix the test, but it should be fixed with latest image.

eplanet (approved):

nextcloud/tests/compose/nextcloud-standalone
hithwen: Usually pinning to `latest` image is a bad idea because then tests break unexpectedly. but you can open a PR later to change it to a newer image

Issue #803, opened 2021-02-10
nvml check spams logs on non-GPU machines.


therc:
On non-GPU machines that lack the libnvidia-ml.so.1 library, the nvml check will fall with a very ugly stack trace that only spams logs.

```
2021-02-10 20:33:54 UTC | CORE | ERROR | (pkg/collector/runner/runner.go:292 in work) | Error running check nvml: [{"message": "NVML Shared Library Not Found", "traceback": "Traceback (most recent call last):
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/pynvml/nvml.py\", line 734, in _load_nvml_library
    nvml_lib = CDLL(\"libnvidia-ml.so.1\")
  File \"/opt/datadog-agent/embedded/lib/python3.8/ctypes/__init__.py\", line 373, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libnvidia-ml.so.1: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/base/checks/base.py\", line 876, in run
    self.check(instance)
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/nvml/nvml.py\", line 81, in check
    with NvmlInit():
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/datadog_checks/nvml/nvml.py\", line 26, in __enter__
    NvmlCheck.N.nvmlInit()
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/pynvml/nvml.py\", line 742, in nvmlInit
    _load_nvml_library()
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/pynvml/nvml.py\", line 736, in _load_nvml_library
    check_return(NVML_ERROR_LIBRARY_NOT_FOUND)
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/pynvml/nvml.py\", line 366, in check_return
    raise NVMLError(ret)
pynvml.nvml.NVMLError_LibraryNotFound: NVML Shared Library Not Found
"}]
```

It's much easier to use the same image and setup on all nodes, regardless of the hardware they might or might not have.
What's the quickest and cleanest way for a custom check to signal at runtime that it should be disabled? Presumably, this would occur in the constructor.

[...]

PR #817, opened 2021-03-03
Make nvml check quieter on non-GPU hosts

therc:
Check nvml library once at instantiation time and skip any further check runs if we failed
Also stop pinning datadog-checks-dev to 3.1.0

### What does this PR do?

This PR makes the check attempt to load libnvml just once, so that we don't fail every 15s on machines without a GPU.

### Motivation

See https://github.com/DataDog/integrations-extras/issues/803

We have many machines with GPUs, but at least as many without. We don't need the spam.

### Review checklist

- [ ] Git history is clean

### Additional Notes

### DIFF
--- a/nvml/datadog_checks/nvml/nvml.py
+++ b/nvml/datadog_checks/nvml/nvml.py
@@ -73,11 +73,30 @@ class NvmlCheck(AgentCheck):
     """Lock for the object known_tags."""
     _thread = None
     """Daemon thread updating k8s tag information in the background."""
+    should_run = False
+    """Whether libnvml can be found and the check can thus run."""
+
+    def __init__(self, name, init_config, instances):
+        super(NvmlCheck, self).__init__(name, init_config, instances)
+        # self.N = pynvml
+        if self.is_nvml_library_available():
+            # Start thread once and keep it running in the background
+            self._start_discovery()
+            self.should_run = True
+
+    def is_nvml_library_available(self):
+        try:
+            self.N.nvmlInit()
+            self.N.nvmlShutdown()
+            return True
+        except pynvml.nvml.NVMLError_LibraryNotFound:
+            self.log.warning("Can't open NVML, is this a GPU host? Turning check off.")
+            return False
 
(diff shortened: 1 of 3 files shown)


therc:
@cep21 this seems to work for me.

therc:
I wish I knew how to increase the check frequency from inside the class itself, but I suspect it's set from the agent side.

nvml/datadog_checks/nvml/nvml.py
```diff
@@ -73,11 +73,30 @@ class NvmlCheck(AgentCheck):
     """Lock for the object known_tags."""
     _thread = None
     """Daemon thread updating k8s tag information in the background."""
+    should_run = False
+    """Whether libnvml can be found and the check can thus run."""
+
+    def __init__(self, name, init_config, instances):
+        super(NvmlCheck, self).__init__(name, init_config, instances)
+        # self.N = pynvml
+        if self.is_nvml_library_available():
```
hithwen, line 82: I think it should log something if its not available (error or debug). On machines where a GPU is not available the check should not be enabled in the first place.

hithwen:
Can you also bump the version and update the changelog to release a new version with this fix? You can use the `ddev release changelog` command for this.

nvml/datadog_checks/nvml/nvml.py
cep21: Looks like this is part of is_nvml_library_available now

cep21 (approved):
LGTM!

hithwen (changes requested):
Can you also bump the version and update the changelog to release a new version with this fix? You can use the ddev release changelog command for this.

hithwen:
For these changes to be released a new PR with a version bump will be needed

srikiz:
@therc - I am trying to use your changes with `agent integration install datadog-nvml==1.0.0 -r -t -v` but I don't see the changes getting reflected.
Is 1.0.0 the correct integration version?  Or I am missing something here?

PR #833, opened 2021-03-26
Make integrations public

hithwen:
So they show in the docs and are easier to discover

### DIFF
--- a/gatekeeper/manifest.json
+++ b/gatekeeper/manifest.json
@@ -19,7 +19,7 @@
     "configuration & deployment"
   ],
   "type": "check",
-  "is_public": false,
+  "is_public": true,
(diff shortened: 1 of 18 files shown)


arapulido (approved):

gatekeeper/manifest.json
```diff
@@ -19,7 +19,7 @@
     "configuration & deployment"
   ],
   "type": "check",
-  "is_public": false,
+  "is_public": true,
```
hithwen, line 22: leave this as false for now to avoid migration issues
```suggestion
  "is_public": false,
```

arapulido (approved):

platinummonkey (approved):

PR #897, opened 2021-05-31
Revert "Revert "Adding JFrog Xray Violations and Logs Integrations""

hithwen:
Reverts DataDog/integrations-extras#896
Fixes will need to be applied here.


kayayarai (approved):
I had to rearrange the structure quite a bit to get it to build for the docs site. I'm not sure what exactly fixed it, so I committed to the branch myself. Could have been: Headings within ordered list, use of three backticks for in-text monospace (should be just one), nested ordered lists, or something else.

PR #900, opened 2021-06-03
Bug fix for getting values from cashe anf url forming for pulsar api 

dblagojevic-daitan:
Bug fix for getting values from cashe anf url forming for pulsar api


dblagojevic-daitan:
Forgot to pull from master first, will submit again

PR #902, opened 2021-06-04
Update Filebeat to use request wrapper

npezzotti:
### What does this PR do?
- Updates Filebeat to use RequestsWrapper
- Modifies `normalize_metrics` logic to use base class namespace formatting
- Updates tests to use instance when instantiating check

### Motivation
- Standardize behavior for HTTP requests
- Leverage standard functionality of base class for formatting metric names

### Review checklist

- [ ] Git history is clean

### Additional Notes

### DIFF
--- a/filebeat/datadog_checks/filebeat/filebeat.py
+++ b/filebeat/datadog_checks/filebeat/filebeat.py
@@ -216,6 +216,11 @@ class FilebeatCheck(AgentCheck):
         AgentCheck.__init__(self, *args, **kwargs)
         self.instance_cache = {}
 
+        # preserve backwards compatible default timeouts
+        if self.instance and self.instance.get('timeout') is None:
+            if self.init_config.get('timeout') is None:
+                self.init_config['timeout'] = 2
+
     def check(self, instance):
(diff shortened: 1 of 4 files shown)


filebeat/datadog_checks/filebeat/filebeat.py
```diff
@@ -175,10 +174,6 @@ def __init__(self, instance):
                 "If given, filebeat's only_metrics must be a list of regexes, got %s" % (self._only_metrics,)
             )
 
-        self._timeout = instance.get("timeout", 2)
```
coignetp, line 178: Let's keep it at 2 seconds in the constructor and override the default value so we keep the same behaviour

coignetp (changes requested):

filebeat/datadog_checks/filebeat/filebeat.py
npezzotti: I added some logic to override the default of 10 set in the RequestWrapper and added a test. I'm not sure if there might be a better way or if the check should just use the request wrappers default but I'm happy to refactor if needed!

coignetp (approved):

Issue #909, opened 2021-06-14
Logstash is missing DLQ metrics


Deroin:
Back on Agent 5.x the logstash integration supplied a metric `logstash.pipeline.dead_letter_queue.queue_size_in_bytes`. With this you could easily monitor whether or not documents get rejected by logstash.

We are now in the process of upgrading to Agent 7.x and noticed, that the integration no longer reports this value. It would be great to get this feature back.

**Output of the [info page](https://docs.datadoghq.com/agent/guide/agent-commands/#service-status)**

```text
===============
Agent (v7.27.0)
===============

  Status date: 2021-06-14 09:01:37.502 UTC (1623661297502)
  Agent start: 2021-06-14 08:39:47.593 UTC (1623659987593)
  Pid: 3.153379e+06
  Go Version: go1.14.12
  Python Version: 3.8.5
  Build arch: amd64
  Agent flavor: agent
  Check Runners: 4
  Log File: /var/log/datadog/agent.log
  Log Level: info

...

    logstash (1.0.0)
    ----------------
      Instance ID: logstash:8bc0d46af7a8cccc [OK]
      Configuration Source: file:/etc/datadog-agent/conf.d/logstash.d/conf.yaml
      Total Runs: 87
      Metric Samples: Last Run: 136, Total: 11,560
      Events: Last Run: 0, Total: 0
      Service Checks: Last Run: 1, Total: 87
      Average Execution Time : 24ms
      Last Execution Date : 2021-06-14 09:01:27 UTC (1623661287000)
      Last Successful Execution Date : 2021-06-14 09:01:27 UTC (1623661287000)
```

hithwen:
Pinging @ervansetiawan as the logstash integration maintainer.
However if you look at the check history https://github.com/DataDog/integrations-extras/commits/master/logstash/datadog_checks/logstash/logstash.py it does not seem like the functionality has changed since it was developed. Also `logstash.pipeline.dead_letter_queue.queue_size_in_bytes` is not present on t[he list of metrics that integration emits](https://github.com/DataDog/integrations-extras/blob/master/logstash/metadata.csv). 

Maybe @ervansetiawan  can help shed some light here

Deroin:
[...]

PR #914, opened 2021-06-16
Add dead letter queue size metric

Deroin:
### What does this PR do?

As discussed in #909 this adds a new metric for dead letter queue size to the logstash integration.

### Motivation

We used this in production before, we're missing it after an infrastructure upgrade and then noticed we never provided it for public use.

### Review checklist

- [ ] Git history is clean

### Additional Notes

I don't know if I should notify the docs team about this or not. Also don't know if you will manage this checklist above or if I should do it. If there needs to be anything done by me just poke me.

I had to adapt the test config for logstash5 by not only enabling dead letter queue but also creating a new input config which keeps logstash running for a bit. On logstash5 the DLQ is started a few seconds after logstash started and with the stdin input logstash stops before that. I also had to increase the wait time for tests due to that fact. On my local machine it passed now, but the wait time may need to be increased again when it fails on build systems. Unfortunately this depends on the speed of the machine executing tests. For logstash 6 and 7 this is no issue as DQL is started at the same time as the main pipeline.

An alternative would be either to not test for the dead letter queue metric or to find a way to exclude it for logstash5 tests.


logstash/tests/docker/logstash5/logstash.yml
```diff
@@ -0,0 +1,200 @@
+# Settings file in YAML
```
coignetp, line 1: I don't think we need all these examples in the config file, let's keep a few lines only

coignetp (changes requested):
Thank you for opening the PR! It looks good to me, I just have one suggestion

Pinging @ervansetiawan as the owner of the check

logstash/tests/docker/logstash5/logstash.yml
ervansetiawan-tr: That makes sense, @coignetp. Other than that, it looks good.

Deroin: Removed all comment lines

coignetp (approved):

hithwen:
Note: to release a new version of this check it needs a version bump with `ddev release changelog`

PR #925, opened 2021-07-05
Pin docker images

hithwen:
Pin docker images to avoid CI from breaking on new releases


coignetp (approved):

hithwen:
Hi @KosukeKamiya we are pinning the testing versions to the ones that were around when this integration was submitted. The reason is that this integration is broken on CI. You can open a PR with a pin to a newer version if you want

PR #931, opened 2021-07-06
Make assertions optional

hithwen:
Not all metrics are emitted on every test run, making the test flake


hithwen:
cc @KosukeKamiya

coignetp (approved):

PR #950, opened 2021-08-05
Add a FoundationDB Integration

jnthn:
### What does this PR do?

Adds a FoundationDB integration. It includes:

* A service check
* Collection of numerous metrics
* Support for collecting metrics from a FoundationDB cluster configured with TLS
* Instructions/suggested configuration for ingesting trace logs
* A range of dashboards
* A range of monitors
* A range of saved views

### Motivation

Being able to monitor FoundationDB using Datadog.

### Review checklist

- [X] PR has a [meaningful title](https://github.com/DataDog/integrations-core/blob/master/CONTRIBUTING.md#pull-request-title) or PR has the `no-changelog` label attached
- [X] Feature or bugfix has tests
- [X] Git history is clean


foundationdb/manifest.json
```diff
@@ -0,0 +1,170 @@
+{
+  "display_name": "FoundationDB",
+  "maintainer": "datadog-integrations@edument.cz",
+  "manifest_version": "1.0.0",
+  "name": "foundationdb",
+  "metric_prefix": "foundationdb.",
+    "metric_to_check": [
```
hithwen, line 7: metric to check should include just one or two metrics. This field is used to display, on the integration tile if the integration is enabled and working properly so it should be a metric that is always emitted.

foundationdb/manifest.json
```diff
@@ -0,0 +1,170 @@
...
+  "short_description": "FoundationDB integration",
+  "guid": "ed46c9de-3a62-46f0-98a6-19344746dd42",
+  "support": "contrib",
+  "supported_os": [
+    "linux",
+    "mac_os",
+    "windows"
+  ],
+  "public_title": "FoundationDB",
+  "categories": [
```
hithwen, line 136:
```suggestion
  "categories": [
    "data store",
  ]
```

foundationdb/assets/configuration/spec.yaml
```diff
@@ -0,0 +1,50 @@
...
+files:
+- name: foundationdb.yaml
+  options:
+  - template: init_config
+    options:
+    - template: init_config/default
+  - template: instances
+    options:
+    - template: instances/default
+    - name: base_command
```
hithwen, line 11: This feels like a bad approach because it requires the agent to run locally on the same machine as foundation db plus having to execute a subrpcess and parse the result is usually error prone so it should only be done when there is no other alternative. Typically integrations use a library or API to communicate with services and in the case of foundation db there seems to be an officially apple supported python library https://apple.github.io/foundationdb/api-python.html.

hithwen (changes requested):

foundationdb/assets/configuration/spec.yaml
ofek: Strongly agree, here is how: https://apple.github.io/foundationdb/mr-status.html

ofek (changes requested):

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+## Setup
+
+Both the check and the metrics apply to the FoundationDB cluster as a whole,
+and so should only be installed on one host (which need not be one that is
+running FoundationDB, but just one with access to it). The host chosen to
+collect the metrics is required to have the [FoundationDB client][8]
+installed. It is possible to monitor multiple FoundationDB clusters if
+desired, by adding multiple instances in the configuration file.
+
+By contrast, logs are written on each machine, and thus for collecting logs it
```
apigirl, line 18:
```suggestion
Since logs are written on each machine, to collect logs, configure configure FoundationDB log collection on every host.
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+is necesary to configure FoundationDB log collection on every host.
```
apigirl, line 19:
```suggestion
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+Follow the instructions below to install and configure this check or log
```
apigirl, line 21:
```suggestion
Follow the instructions below to install and configure the check or log
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+the root of your Agent's configuration directory to start collecting
+FoundationDB metrics. See the [sample foundationdb.d/conf.yaml][3] for
+all available configuration options.
+
+The details of the cluster to check are found under the `instances`
+section. Some hints:
+
+* The `base_command` is correct by default if the `fdbcli` command is in
+  your `PATH` and the FoundationDB instance is on the current machine.
+  If you have `fdbcli` installed in a container, then you could modify
```
apigirl, line 56:
```suggestion
  If you have `fdbcli` installed in a container, then modify
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+  this command to execute it inside of the container.
+* The `cluster_file` is where `fdbcli` finds the cluster connection
+  details. If this is not specified, then it will be searched for in
+  the [default location][10]. Ensure that your cluster file is in
+  that location, and set this property if it is located elsewhere.
+
+To check more than one FoundationDB cluster, add an extra item to the
+`instances` list. In this sitaution, at least one of the instances must
```
apigirl, line 64:
```suggestion
`instances` list. In this situation, at least one of the instances must
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+FoundationDB writes XML logs by default, however the Datadog integration
+expects JSON logs instead. Thus a configuration change shall be needed to
+FoundationDB itself first.
+
+1. Locate your `foundationdb.conf` file. Under the `fdbserver` section, add
+   or change the key `trace_format` to have the value `json`. Also make a
+   note of the `logdir`.
+
+```
```
apigirl, line 83:
```suggestion
    ```
```

apigirl: any examples or notes under a step in an ordered list must be indented or spaced or the docs site will break the list and restart with 1 for the next list item

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+[fdbserver]
```
apigirl, line 84:
```suggestion
    [fdbserver]
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+...
```
apigirl, line 85:
```suggestion
    ...
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+logdir = /var/log/foundationdb
```
apigirl, line 86:
```suggestion
    logdir = /var/log/foundationdb
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+trace_format = json
```
apigirl, line 87:
```suggestion
    trace_format = json
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+```
```
apigirl, line 88:
```suggestion
    ```
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+2. Restart the FoundationDB server so the changes take effect. Verify that
+   logs in the logdir are now being written in JSON.
+
+2. Ensure that log collection in the Datadog Agent is enabled. In your
```
apigirl, line 93:
```suggestion
3. Ensure that log collection in the Datadog Agent is enabled. In your
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+   `datadog.yaml` file, make sure this appears:
+
+```yaml
```
apigirl, line 96:
```suggestion
    ```yaml
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+logs_enabled: true
```
apigirl, line 97:
```suggestion
    logs_enabled: true
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+```
```
apigirl, line 98:
```suggestion
    ```
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+3. In the `foundationdb.d/conf.yaml` file, uncomment the `logs` section
```
apigirl, line 100:
```suggestion
4. In the `foundationdb.d/conf.yaml` file, uncomment the `logs` section
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+   and set the path to that found in your FoundationDB configuration file,
+   appending `*.json`.
+
+```yaml
```
apigirl, line 104:
```suggestion
    ```yaml
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+logs:
```
apigirl, line 105:
```suggestion
    logs:
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+  - type: file
```
apigirl, line 106:
```suggestion
      - type: file
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+    path: /var/log/foundationdb/*.json
```
apigirl, line 107:
```suggestion
        path: /var/log/foundationdb/*.json
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+    service: foundationdb
```
apigirl, line 108:
```suggestion
        service: foundationdb
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+    source: foundationdb
```
apigirl, line 109:
```suggestion
        source: foundationdb
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+```
```
apigirl, line 110:
```suggestion
    ```
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+4. Make sure the Datadog agent has the privileges required to list the
```
apigirl, line 112:
```suggestion
5. Make sure the Datadog Agent has the privileges required to list the
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+   directory and read its files.
+
+5. Restart the Datadog agent.
```
apigirl, line 115:
```suggestion
5. Restart the Datadog Agent.
```

apigirl (commented):
Documentation review! I am not sure if it is your text editor you are using, but in addition to this feedback, please remove the forced line breaks to format the readme file. Our docs site and the tile in the app handles line breaks in a screen size responsive way, so based on what sort of device you are viewing on. So, for example, you'll want to make sure sentences and paragraphs are all on one line instead of having a return in the middle of them. Use return line breaks for if you are wanting to start a new paragraph, example or section.

foundationdb/assets/configuration/spec.yaml
jnthn: > This feels like a bad approach because it requires the agent to run locally on the same machine as foundation db

This is incorrect; it requires the FoundationDB *client binaries* to be installed on the machine that the agent runs on. However, the Python library also requires these very same client binaries. It seems the library and CLI are distributed together, so if you have the binaries that the Python library wants installed, you have the CLI too. Thus using the Python library means the integration would have more dependencies, not less. The link to the Python library you gave also notes that the additional issue of having to make sure `libffi` is installed too depending on Python version:

> The FoundationDB Python API is compatible with Python 2.7 - 3.7. You will need to have a Python version within this range on your system before the FoundationDB Python API can be installed. Also please note that Python 3.7 no longer bundles a full copy of libffi, which is used for building the _ctypes module on non-macOS UNIX platforms. Hence, if you are using Python 3.7, you should make sure libffi is already installed on your system.

Also, about this:

> having to execute a subrpcess and parse the result is usually error prone

While I'd certainly agree in the case of parsing text output that could change at any time, here the output is JSON - in fact, the very same we'd get if we used the Python library and queried the special key.

We can change this if you really insist, but it's the way it is not because we were unaware of the Python binding option, but because there seems to be no advantages to be had in return for additional moving parts.

jnthn:
> Documentation review! 

Thanks for all of the corrections; I'll apply them. I'm a bit confused about this:

> I am not sure if it is your text editor you are using, but in addition to this feedback, please remove the forced line breaks to format the readme file.

So far as I'm aware, all Markdown rendering engines disregard single line breaks (only two or more are significant, and start a new paragraph; a single line break maps into a space). Thus their presence should have no impact. (I assume that Markdown was designed this way because many programming editors do not wrap long lines.)

foundationdb/README.md
jnthn: Duplicated word "configure" in the correction

jnthn: Applied manually with the dupe word removed.

foundationdb/manifest.json
jnthn: Applied, though without the trailing comma (sigh, JSON :-)).

foundationdb/manifest.json
jnthn: Makes sense; changed.

foundationdb/assets/configuration/spec.yaml
ofek: FYI the Agent is guaranteed to have `libffi`. Also, is Python 3.7 a hard upper limit or just what they've tested?

jnthn: > FYI the Agent is guaranteed to have libffi

OK, that's at least something if we have to go the Python way.

> Also, is Python 3.7 a hard upper limit or just what they've tested?

I've not found confirmation either way, although that line in the documentation was last edited in March 2019, so it's most likely out-dated.

ofek:
@apigirl yeah we also use line breaks in -core. is that a new requirement?

foundationdb/assets/configuration/spec.yaml
hithwen: Adding coding dependencies is not an issue. Adding dependencies that the customer has to manually install is not a great experience for the user so it should be avoided. Executing subprocesses is a source of problems and should be avoided when possible

jnthn: > Adding coding dependencies is not an issue. Adding dependencies that the customer has to manually install is not a great experience for the user so it should be avoided

What I'm trying (and apparently failing) to explain is that it's not a choice of code (I assume we mean Python) dependencies *or* the FoundationDB client binaries. If one creates a clean environment (I used an Ubuntu container), does `pip install foundationdb`, and then tries to connect to a cluster, the result is:

```
Traceback (most recent call last):
  File "x.py", line 2, in <module>
    fdb.api_version(630)
  File "/usr/local/lib/python3.8/dist-packages/fdb/__init__.py", line 68, in api_version
    import fdb.impl
  File "/usr/local/lib/python3.8/dist-packages/fdb/impl.py", line 1300, in <module>
    raise Exception("Unable to locate the FoundationDB API shared library!")
Exception: Unable to locate the FoundationDB API shared library!
```

That is to say, the FoundationDB client binaries also have to be installed **regardless**. I don't believe there's a way to avoid that manual step. Using the Python library does not eliminate it, as demonstrated.

> Executing subprocesses is a source of problems and should be avoided when possible

I believe the concrete points so far have been:

* Parsing process output is fragile. This is true if the output is not meant to be machine readable. But we get JSON back; the command we're invoking was designed for obtaining machine readable status information.
* Making the user install the FoundationDB client is not ideal. This is true, but the client needs to be installed anyway because the Python binding for FoundationDB relies on it.

I'm getting the impression that no matter what arguments I put forward, there's no way this integration will get merged unless we reimplement it using the FoundationDB Python library to do the status query. If that's the case, please say so, and we'll do it; it just seems to me like a net increase in complexity (more dependencies) and things that might go wrong (such as a future incompatibility between the FoundationDB Python binding bundled in the integration and the installed client libraries).

ofek: Can queries be made by via the CLI? For our other database integrations we allow users to define custom queries to monitor.

ofek: We've decided to begin shipping the `foundationdb` Python library with the Agent. As far as [API versioning](https://apple.github.io/foundationdb/api-python.html#api-versioning) goes, let's add an option that defaults to the current `630`.

I'll update here when I open the dependency PR.

ofek: Thanks for your patience here! We've added a way ship dependencies for this use case: https://github.com/DataDog/integrations-core/pull/10040

1. Nope, no need. If you ever wish to change the version, modify these files: https://github.com/DataDog/integrations-core/pull/10050
2. Yes, set [this](https://github.com/edumentab/integrations-extras/blob/bf5c223ae671fcac6b284feaf9597c04bb1cb63b/foundationdb/setup.py#L28) to `21.1.0`.
3. For regular tests you'll need to add the dependency to the `requirements-dev.txt` like [this](https://github.com/DataDog/integrations-core/blob/master/ibm_db2/requirements-dev.txt) and I suppose you'll need to create a [script](https://datadoghq.dev/integrations-core/meta/ci/#scripts) in this repo that installs the client ([example](https://github.com/DataDog/integrations-core/blob/master/.azure-pipelines/scripts/ibm_mq/linux/55_install_client.sh)). For tests [marked as E2E](https://datadoghq.dev/integrations-core/ddev/plugins/#agent-check-runner) you'll need to install the client inside of the Agent container using [environment metadata](https://datadoghq.dev/integrations-core/ddev/plugins/#environment-manager) like [this](https://github.com/DataDog/integrations-core/blob/e08d2911e65f0bd3d8b9a7e0566aadae534c94c2/ibm_mq/tests/conftest.py#L216) defined [here](https://github.com/DataDog/integrations-core/blob/e08d2911e65f0bd3d8b9a7e0566aadae534c94c2/ibm_mq/tests/common.py#L140-L144).

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+and, optionally, FoundationDB transaction logs.
+
+## Setup
+
+Both the check and the metrics apply to the FoundationDB cluster as a whole,
+and so should only be installed on one host (which doesn't need to be one that is
+running FoundationDB, but just one with access to it). The host chosen to
+collect the metrics is required to have the [FoundationDB client][8]
+installed. It is possible to monitor multiple FoundationDB clusters
+by adding multiple instances in the configuration file.
```
apigirl, line 16:
```suggestion
Both the check and the metrics apply to the FoundationDB cluster as a whole, and so should only be installed on one host (which doesn't need to be one that is running FoundationDB, but just one with access to it). The host chosen to collect the metrics is required to have the [FoundationDB client][8] installed. It is possible to monitor multiple FoundationDB clusters by adding multiple instances in the configuration file.
```

foundationdb/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+Edit the `foundationdb.d/conf.yaml` file, in the `conf.d/` folder at
+the root of your Agent's configuration directory to start collecting
+FoundationDB metrics. See the [sample foundationdb.d/conf.yaml][3] for
+all available configuration options.
+
+The details of the cluster to check are found under the `instances`
+section. Some hints:
+
+* The `base_command` is correct by default if the `fdbcli` command is in
```
ofek, line 54: Let's update to reflect the new implementation

foundationdb/assets/configuration/spec.yaml
```diff
@@ -0,0 +1,39 @@
+name: FoundationDB
+files:
+- name: foundationdb.yaml
+  options:
+  - template: init_config
+    options:
+    - template: init_config/default
+  - template: instances
+    options:
+    - template: instances/default
```
ofek, line 10: Let's put this after check-specific options

foundationdb/datadog_checks/foundationdb/check.py
```diff
@@ -0,0 +1,219 @@
+from datadog_checks.base import AgentCheck
+import json
+from typing import Any
+import fdb
+
+fdb.api_version(600)
```
ofek, line 6: WDYTA making this configurable within `init_config`?

foundationdb/datadog_checks/foundationdb/check.py
```diff
@@ -0,0 +1,219 @@
...
+
+class FoundationdbCheck(AgentCheck):
+    def __init__(self, name, init_config, instances):
+        super(FoundationdbCheck, self).__init__(name, init_config, instances)
+        self.hasDB = False
```
ofek, line 11: WDYTA caching? https://github.com/DataDog/integrations-core/pull/8083

e.g.

```python

    def __init__(self, name, init_config, instances):
        super(FoundationdbCheck, self).__init__(name, init_config, instances)
        ...
        self.db = None

    @property
    def db(self):
        if self.db is None:
            self.db = ...

        return self.db
```

foundationdb/tox.ini
```diff
@@ -0,0 +1,24 @@
...
+basepython = py38
+envlist =
+    py{27,38}
+
+[testenv]
+ensure_default_envdir = true
+envdir =
+    py27: {toxworkdir}/py27
+    py38: {toxworkdir}/py38
+dd_check_style = false
```
ofek, line 13: Let's enable this

foundationdb/tox.ini
```diff
@@ -0,0 +1,24 @@
+[tox]
+minversion = 2.0
+skip_missing_interpreters = true
+basepython = py38
+envlist =
+    py{27,38}
```
ofek, line 6:
```suggestion
    py38
```

foundationdb/tox.ini
```diff
@@ -0,0 +1,24 @@
...
+
+[testenv]
+ensure_default_envdir = true
+envdir =
+    py27: {toxworkdir}/py27
+    py38: {toxworkdir}/py38
```
ofek, line 12:
```suggestion
```

ofek (changes requested):
🥇

foundationdb/datadog_checks/foundationdb/check.py
jnthn: Quoting the FoundationDB docs:

> The purpose of this design is to allow the server, client libraries, or bindings to be upgraded without having to modify application code. 

So no, this shouldn't be configurable, because it's about the API that we're using to implement the integration, and we'd have to review and maybe change the Python code here in the event we did want to increment this in the future.

foundationdb/tox.ini
jnthn: If we must, than at the very end, as the last step before the merge.

foundationdb/tox.ini
jnthn: Hm, does this mean we're dropping support for 2.7? (Fine for us, just want to understand what it's doing.)

foundationdb/README.md
jnthn: Done, and taken a pass over the README as a whole to see if anything else was needed.

foundationdb/assets/configuration/spec.yaml
jnthn: Not sure I follow; you mean that this line (with the `template` key) should move to the end of the file?

foundationdb/tox.ini
jnthn: Ah, and to add *why*, since it could be useful feedback: when using the `ddev` command to run tests, it also would run the style check. When one is at the point of trying to debug a test failure, especially when new to building an integration and with a lot to figure out, about the last thing wanted is "oh, you have a couple of lines that are too many characters long!"

foundationdb/assets/configuration/spec.yaml
ofek: Yes

```yaml

  - template: instances
    options:
    - name: cluster_file
      description: |
        A path to the cluster file specifing connection to the FoundationDB.
        If it is not passed, the check will rely on the binding library to find
        the file using standard locations (e.g. `/etc/foundationdb/fdb.cluster` or
        `/etc/foundationdb/fdb.cluster`).
      value:
        example: /var/lib/foundationdb/fdb.cluster
        type: string
    ...
    - template: instances/default
```

foundationdb/tox.ini
ofek: Yup! We only require Python 3 for this.

foundationdb/tox.ini
FlorianVeaux: Just FYI, you can run `ddev test foundationdb:py38` to only run py38 test without having to care about style test during the development phase. But no blocker to turn it on only once the implementation is fully ready.

jnthn: Ah, good to know; thanks.

foundationdb/assets/configuration/spec.yaml
jnthn: Thanks; done.

foundationdb/datadog_checks/foundationdb/check.py
jnthn: Hm, we did something like this but now I'm reviewing it and I'm confused how the Oracle one you linked is correct. It's constructed like this:

```
    def __init__(self, name, init_config, instances):
        super(Oracle, self).__init__(name, init_config, instances)
```

That is, with many instances, but then goes on to do:

```
        self._server = self.instance.get('server')
        self._user = self.instance.get('user')
```

But...which instance? If we were just constructed with all of them, how can we assume there's one of them?

jnthn: Ah, OK, good. Then I'm comfortable what we're doing (caching the connection on the instance) will work.

foundationdb/datadog_checks/foundationdb/check.py
```diff
@@ -0,0 +1,285 @@
...
+            self.gauge("foundationdb.degraded_processes", cluster["degraded_processes"])
+            degraded_processes = cluster["degraded_processes"]
+
+        if degraded_processes > 0:
+            self.service_check("foundationdb.can_connect", AgentCheck.WARNING, message="There are degraded processes")
+        else:
+            self.service_check("foundationdb.can_connect", AgentCheck.OK)
+
+    def maybe_gauge(self, metric, obj, key, tags=None):
+        if key in obj:
```
yzhan289, line 269: Just wondering, in what cases would this not be the case?

foundationdb/assets/configuration/spec.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: cluster_file
+      description: |
+        A path to the cluster file specifing connection to the FoundationDB.
+        If it is not passed, the check will rely on the binding library to find
+        the file using standard locations (e.g. `/etc/foundationdb/fdb.cluster` or
+        `/etc/foundationdb/fdb.cluster`).
+      value:
+        example: /var/lib/foundationdb/fdb.cluster
+        type: string
+    - name: tls_certificate_file
```
yzhan289, line 19: Have you considered using our existing `http` template? We already have these config options in it: https://github.com/DataDog/integrations-core/blob/master/datadog_checks_dev/datadog_checks/dev/tooling/templates/configuration/instances/http.yaml#L240

jnthn: Using in what way? You mean matching the same keys? That'd be one consistency (across Datadog integrations); we've instead gone with consistency with how these options are named in FoundationDB.

PR #981, opened 2021-09-02
Do not use legacy signature

hithwen:


hithwen:
Pinging @pault84 as codeowner

ChristineTChen (approved):

PR #990, opened 2021-09-08
send filebeat instance tags

OpeOnikute:
### What does this PR do?

Send the filebeat instance tags in the service and metric reporting.
 
### Motivation

The check doesn't recognise the tags set. It sets only one tag in the case of the http service check: `stats_endpoint`, which is set in the code. This makes it hard to create different monitors based on unique tags.

### Review checklist

- [x] Git history is clean

### Additional Notes


filebeat/datadog_checks/filebeat/filebeat.py
```diff
@@ -221,11 +221,14 @@ class FilebeatCheck(AgentCheck):
...
-        # preserve backwards compatible default timeouts
-        if self.instance and self.instance.get('timeout') is None:
-            if self.init_config.get('timeout') is None:
-                self.instance['timeout'] = 2
+        if self.instance:
+            self.tags = self.instance.get('tags', [])
+            # preserve backwards compatible default timeouts
+            if self.instance.get('timeout') is None:
+                if self.init_config.get('timeout') is None:
+                    self.instance['timeout'] = 2
```
coignetp, line 231:
```suggestion
        self.tags = self.instance.get('tags', [])
        # preserve backwards compatible default timeouts
        if self.instance.get('timeout') is None:
            if self.init_config.get('timeout') is None:
                self.instance['timeout'] = 2
```
We don't need to check that actually

coignetp (changes requested):
Just a tiny suggestion

Pinging @wk8 as integration maintainer

filebeat/datadog_checks/filebeat/filebeat.py
OpeOnikute: @coignetp i did that initially, but that causes the test for invalid configs to fail. Makes sense because if an empty config is passed, `self.instance` does not exist.

coignetp: It does make sense, but it should not happen based on how integrations instances are created. However, we can keep it as it only happens in the constructor, thank you for your investigation

coignetp (approved):

PR #1115, opened 2022-01-03
Setting up codescanning for integrations-extras

ganeshkumarsv:
Static code analysis on integrations-extras using Github's Code Scanning.

This PR has Github workflow configs to trigger the Code Scanning on every PR and on push to master branch. As of now only default security based CodeQL queries are configured. We can gradually add queries for Code Quality later. It is a part of Datadog compliance requirements to regularly perform code-scanning on all the customer installed code. As the repository is small, no security alerts were found during the scan.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'python', 'ruby' ]
```
yzhan289, line 22: This repo should only have python files
```suggestion
        language: [ 'python' ]
```

yzhan289 (changes requested):
Thanks for the PR! Can you also add something similar to https://github.com/DataDog/integrations-core/pull/10963/files#diff-5f817878eaf9204edc264b31f944926a01e44b11f34131257be3d84c20f8f51fR3 for `paths-ignore`?

ganeshkumarsv:
@yzhan289 No alerts have been detected on the scan that we ran, so I don't think `paths-ignore` is necessary right now. In case if we notice some alerts coming from any of the test or lib directories, we can add them later.

yzhan289 (approved):

PR #1155, opened 2022-02-15
Add `pyproject.toml` file

ofek:
### Motivation

### Additional Notes


eventstore/pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+readme = "README.md"
+license = "BSD-3-Clause"
+keywords = [
+    "datadog",
+    "datadog agent",
+    "datadog check",
+    "eventstore",
+]
+authors = [
+    { email = "@xorima" },
```
sarah-witt, line 20:
```suggestion
    { email = "jason.field@calastone.com" },
```

PR #1157, opened 2022-02-15
Add `pyproject.toml` file

ofek:
### Motivation

### Additional Notes

### DIFF
--- a/flume/pyproject.toml
+++ b/flume/pyproject.toml
@@ -17,7 +17,7 @@ keywords = [
     "flume",
 ]
 authors = [
-    { name = "Datadog", email = "packages@datadoghq.com" },
+    { email = "kealan.maas@datadoghq.com" },
 ]
 classifiers = [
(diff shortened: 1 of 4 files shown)


flume/pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Intended Audience :: System Administrators",
+    "License :: OSI Approved :: BSD License",
+    "Programming Language :: Python :: 2.7",
+    "Programming Language :: Python :: 3.8",
+    "Topic :: System :: Monitoring",
+]
+dependencies = [
+    "datadog-checks-base",
```
fanny-jiang, line 32: Is it a problem that the base check isn't pinned to a version?

ofek: Not for this afaict

PR #1169, opened 2022-02-15
Add `pyproject.toml` file

ofek:
### Motivation

### Additional Notes


neutrona/pyproject.toml
```diff
@@ -0,0 +1,64 @@
...
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Intended Audience :: System Administrators",
+    "License :: OSI Approved :: BSD License",
+    "Programming Language :: Python :: 2.7",
+    "Programming Language :: Python :: 3.8",
+    "Topic :: System :: Monitoring",
+]
+dependencies = [
+    "datadog-checks-base",
```
yzhan289, line 32: Is this fine not having a minimum version?

ofek: Yup!

PR #1193, opened 2022-02-15
Add `pyproject.toml` file

ofek:
### Motivation

### Additional Notes


platinummonkey (approved):

ofek:
Fixed by https://github.com/DataDog/integrations-core/pull/11516

PR #1328, opened 2022-06-01
Move v2 manifest field `classifier_tags` under `tile`

ofek:
### Motivation

https://github.com/DataDog/integrations-core/pull/12122


joneshf-dd (approved):
If I'm understanding the linked PR correctly, all that this PR does is move the `classifier_tags` from the top-level down into the `tile` field.

I'm not sure what the ramifications of this change are, so someone with more knowledge should probably weigh in here. But from the perspective of the files the Apps SDK team is responsible for, this seems like the right change!

PR #1541, opened 2022-09-19
Add integration for collecting Go profiles

nsrip-dd:
### What does this PR do?

This PR adds an integration, tentatively called `go_pprof_scraper`, which collects runtime performance profiles from Go programs via the Go standard library's [`/debug/pprof`](https://pkg.go.dev/net/http/pprof) endpoint every minute and uploads them to Datadog through the agent's profile proxy.

### Motivation

Many programs written in Go, such as etcd, the Kubernetes components, and Consul make Go runtime profiles available through an endpoint provided by the standard library. These programs can have performance problems which profiles help to diagnose, and Datadog's profiling UI makes it easy to explore and compare profiles.

Normally, we encourage Datadog customers to instrument their Go apps using our profiler client library. However, modifying external applications can be impractical. This PR adds a way for customers to get profiles from these services into our UI.

### Review checklist

- [ ] Git history is clean

### Additional Notes

This PR has end-to-end tests, and I have also run it locally to upload profiles to Datadog and confirmed that it works as intended. There are a few remaining TODOs, mostly due to me not knowing which APIs to use to get certain information (container ID tags, accurate agent configuration).


go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
```diff
@@ -0,0 +1,208 @@
...
+                    "{} is not a valid profile type, must be one of {}".format(profile, VALID_PROFILE_TYPES)
+                )
+        self.profiles = list(set(self.profiles))
+
+        self.cumulative = self.instance.get("cumulative", True)
+        self.tags = self.instance.get("tags", [])
+        self.tags.extend(self.init_config.get("tags", []))
+
+    def _get_apm_config(self):
+        # TODO: for some reason the config I get here doesn't match the real
```
nsrip-dd, line 81: This is probably the biggest issue I've found so far. If I start the agent without explicitly enabling APM, The `datadog_agent.get_config("apm_config")` call will still return a blob of configuration that says APM is enabled. But then the check subsequently fails to upload profiles to the profile proxy since it's not actually running.

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
```diff
@@ -0,0 +1,208 @@
...
+            add_form_field("end", _timeformat(start + datetime.timedelta(seconds=self.duration)))
+            add_form_field("tags[]", "runtime:go")
+            add_form_field("tags[]", "service:{}".format(self.service))
+            add_form_field("tags[]", "runtime-id:{}".format(self.runtime_id))
+            if self.env:
+                add_form_field("tags[]", "env:{}".format(self.env))
+            for tag in self.tags:
+                add_form_field("tags[]", tag)
+
+            # TODO: container ID? If we can get it, it should be added as a
```
nsrip-dd, line 174: Is there a Python API for getting other tags associated with the service being checked? Specifically if this is running on a cluster using autodiscovery, and if there's a container ID tag available.

go_pprof_scraper/datadog_checks/go_pprof_scraper/data/conf.yaml.example
```diff
@@ -0,0 +1,82 @@
...
+    #   - <KEY_2>:<VALUE_2>
+
+## Every instance is scheduled independent of the others.
+#
+instances:
+
+    ## @param service_name - string - required
+    ## Name of the service, which will be tagged as "service:<SERVICE>" on uploaded profiles
+    #
+  - service_name: myservice
```
nsrip-dd, line 21: If I call this `service`, the unit tests fail on the `assert_service_check` calls with an error saying that a `service` tag isn't allowed. I saw I can set a variable somewhere to make this error go away, but is that the appropriate thing to do?

go_pprof_scraper/README.md
```diff
@@ -0,0 +1,67 @@
...
+**NOTE**: Using this integration will result in billing for hosts under Datadog's [Continuous Profiler][10] service.
+For more on Continuous Profiler billing, consult our [billing documentation][11].
+
+## Setup
+
+Follow the instructions below to install and configure this check for an Agent running on a host. For containerized environments, see the [Autodiscovery Integration Templates][3] for guidance on applying these instructions.
+
+### Installation
+
+TODO: update this
```
nsrip-dd, line 20: What does this need to be for an integration in `integration-extras`?

FlorentClarret:
Side note: The `validations` job will be fixed on the next run, we fixed the `ddev validate ci` command, nothing to do on your side

go_pprof_scraper/datadog_checks/go_pprof_scraper/data/conf.yaml.example
FlorentClarret: `service` is a reserved parameter. You should use the `default` template to get it and add your other parameters to the `spec.yaml` file. You can find an example [here](https://github.com/DataDog/integrations-core/blob/master/airflow/assets/configuration/spec.yaml), which generates [this file](https://github.com/DataDog/integrations-core/blob/master/airflow/datadog_checks/airflow/data/conf.yaml.example)

go_pprof_scraper/README.md
FlorentClarret: Something like that: https://github.com/DataDog/integrations-extras/tree/master/redisenterprise#installation

The only difference between a -core and an -extras integration is that -extras integrations need to be installed manually

go_pprof_scraper/README.md
```diff
@@ -0,0 +1,67 @@
...
+Need help? Contact [Datadog support][9].
+
+
+[1]: https://pkg.go.dev/net/http/pprof
+[2]: https://app.datadoghq.com/account/settings#agent
+[3]: https://docs.datadoghq.com/agent/kubernetes/integrations/
+[4]: https://github.com/DataDog/integrations-core/blob/master/check/datadog_checks/check/data/conf.yaml.example
+[5]: https://docs.datadoghq.com/agent/guide/agent-commands/#start-stop-and-restart-the-agent
+[6]: https://docs.datadoghq.com/agent/guide/agent-commands/#agent-status-and-information
+[7]: https://github.com/DataDog/integrations-core/blob/master/check/metadata.csv
```
FlorentClarret, line 61:
```suggestion
[7]: https://github.com/DataDog/integrations-extras/blob/master/go_pprof_scraper/metadata.csv
```

go_pprof_scraper/README.md
```diff
@@ -0,0 +1,67 @@
...
+[8]: https://github.com/DataDog/integrations-core/blob/master/check/assets/service_checks.json
```
FlorentClarret, line 62:
```suggestion
[8]: https://github.com/DataDog/integrations-extras/blob/master/go_pprof_scraper/assets/service_checks.json
```

go_pprof_scraper/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Validation
+
+[Run the Agent's status subcommand][6] and look for `go_pprof_scraper` under the Checks section.
+
+## Data Collected
+
+### Metrics
+
+See [metadata.csv][7] for a list of metrics provided by this integration.
```
FlorentClarret, line 38: Looks like your csv file is empty. In this case you can say that this integration does not collect metrics

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
FlorentClarret: [This](https://github.com/DataDog/integrations-extras/pull/1541/files#diff-6321a81caa84fe8dfa36d5a3c1be0d1d118eeaebfd4fda66a530b2334c060da2R11-R18) is the config you get when running the test env because we use [this](https://github.com/DataDog/integrations-extras/pull/1541/files#diff-6321a81caa84fe8dfa36d5a3c1be0d1d118eeaebfd4fda66a530b2334c060da2R22-R37) to start it

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,35 @@
+{
+  "display_name": "Go-pprof-scraper",
+  "maintainer": "help@datadoghq.com",
+  "manifest_version": "1.0.0",
```
FlorentClarret, line 4: You need to update your `ddev` command to the latest version. We do not support the v1 anymore, it should generate you a v2 file now. Alternatively you could use https://datadoghq.dev/integrations-core/ddev/cli/#ddev-meta-manifest-migrate

go_pprof_scraper/README.md
nsrip-dd: Updated

go_pprof_scraper/README.md
nsrip-dd: Got it, thanks! I updated the installation guide to follow that example.

go_pprof_scraper/datadog_checks/go_pprof_scraper/data/conf.yaml.example
nsrip-dd: As background, for profiling we always want to have a `service` tag for uploaded profiles, so I want to make it required for configuring this integration. In my tests I add the service value to the config. If I add `service` to the config, though, I get an error like this when I run the end-to-end tests:

```
E   Exception: Metric go_pprof_scraper.can_connect was submitted with a forbidden tag: service. Please rename this tag, or skip the tag validation with DDEV_SKIP_GENERIC_TAGS_CHECK environment variable.
```

I've changed the config to use the `default` template anyway. Should I just add the `DDEV_SKIP_GENERIC_TAGS_CHECK` environment variable like this error message says?

nsrip-dd: I decided to skip the generic tags check.

.github/CODEOWNERS
```diff
@@ -53,6 +53,7 @@
 /fluxcd/                                 @melchiormoulin melchior.moulin@gmail.com
 /gatekeeper/                             @arapulido ara.pulido@datadoghq.com
 /gitea/                                  @FlorentClarret
+/go_pprof_scraper/                       @nsrip-dd
```
FlorentClarret, line 56: Could you also update this section so the doc team will keep an eye on the docs? https://github.com/DataDog/integrations-extras/blob/master/.github/CODEOWNERS#L167

go_pprof_scraper/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Troubleshooting
+
+Need help? Contact [Datadog support][9].
+
+
+[1]: https://pkg.go.dev/net/http/pprof
+[2]: https://app.datadoghq.com/account/settings#agent
+[3]: https://docs.datadoghq.com/agent/kubernetes/integrations/
+[4]: https://github.com/DataDog/integrations-core/blob/master/check/datadog_checks/check/data/conf.yaml.example
```
FlorentClarret, line 66:
```suggestion
[4]: https://github.com/DataDog/integrations-extras/blob/master/go_pprof_scraper/datadog_checks/go_pprof_scraper/data/conf.yaml.example
```

go_pprof_scraper/datadog_checks/go_pprof_scraper/__about__.py
```diff
@@ -0,0 +1,4 @@
+# (C) Datadog, Inc. 2022-present
+# All rights reserved
+# Licensed under a 3-clause BSD style license (see LICENSE)
+__version__ = '0.0.1'
```
FlorentClarret, line 4: Do you want to keep 0.0.1 or have 1.0.0 directly once merged?

go_pprof_scraper/assets/service_checks.json
```diff
@@ -0,0 +1 @@
+[]
```
FlorentClarret, line 1: The dasbhoard file is empty (can't comment on it so commenting here)

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
FlorentClarret: @ofek any idea?

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
```diff
@@ -0,0 +1,208 @@
...
+        response.raise_for_status()
+        if profile == "cpu" or self.cumulative:
+            name = profile
+        else:
+            name = "delta-{}".format(profile)
+        return name, response
+
+    def check(self, _):
+        if not self.apm_enabled:
+            self.service_check("can_connect", AgentCheck.CRITICAL, message="the trace agent is not enabled")
```
FlorentClarret, line 141: This check should be declared in the `service_checks.json` file

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,46 @@
+{
+  "manifest_version": "2.0.0",
+  "app_uuid": "2b13f6b1-d3ba-4254-93c0-2ceaf783cd85",
+  "app_id": "go-pprof-scraper",
+  "display_on_public_website": false,
```
FlorentClarret, line 5: Do you want this integration to be "private" for now?

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,46 @@
...
+  "tile": {
+    "overview": "README.md#Overview",
+    "configuration": "README.md#Setup",
+    "support": "README.md#Support",
+    "changelog": "CHANGELOG.md",
+    "description": "",
```
FlorentClarret, line 11: Could you also add a description? It will be displayed on the UI

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,46 @@
...
+      "configuration": {
+        "spec": "assets/configuration/spec.yaml"
+      },
+      "events": {
+        "creates_events": false
+      },
+      "metrics": {
+        "prefix": "go_pprof_scraper.",
+        "check": [],
+        "metadata_path": "metadata.csv"
```
FlorentClarret, line 39: Do you plan to add metrics in the future? 

If not, you can remove this section and the `metadata.csv`

FlorentClarret (changes requested):
Sorry for the delay. Left a couple of comments and pinged ofek to take a look at your outstanding questions

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,46 @@
+{
+  "manifest_version": "2.0.0",
+  "app_uuid": "2b13f6b1-d3ba-4254-93c0-2ceaf783cd85",
+  "app_id": "go-pprof-scraper",
```
FlorentClarret, line 4:
```suggestion
  "app_id": "go_pprof_scraper",
```

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,46 @@
...
+  "display_on_public_website": false,
+  "tile": {
+    "overview": "README.md#Overview",
+    "configuration": "README.md#Setup",
+    "support": "README.md#Support",
+    "changelog": "CHANGELOG.md",
+    "description": "",
+    "title": "Go-pprof-scraper",
```
FlorentClarret, line 12:
```suggestion
    "title": "Go pprof scraper",
```

That's what people will see in the DD UI

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,46 @@
...
+  "author": {
+    "support_email": "help@datadoghq.com",
+    "name": "Datadog",
+    "homepage": "https://www.datadoghq.com",
+    "sales_email": "info@datadoghq.com"
+  },
+  "oauth": {},
+  "assets": {
+    "integration": {
+      "source_type_name": "Go-pprof-scraper",
```
FlorentClarret, line 29:
```suggestion
      "source_type_name": "go_pprof_scraper",
```

go_pprof_scraper/manifest.json
```diff
@@ -0,0 +1,46 @@
...
+      "metrics": {
+        "prefix": "go_pprof_scraper.",
+        "check": [],
+        "metadata_path": "metadata.csv"
+      },
+      "service_checks": {
+        "metadata_path": "assets/service_checks.json"
+      }
+    }
+  }
```
FlorentClarret, line 45: Note: If the dashboard was not supposed to be empty, you should also add it here, otherwise just remove the json file

FlorentClarret (changes requested):

go_pprof_scraper/manifest.json
nsrip-dd: Done, I removed the file.

.github/CODEOWNERS
nsrip-dd: Done

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
nsrip-dd: Done. What should I set the "agent_version" field to, though? The latest one? I just copied another file for now.

go_pprof_scraper/datadog_checks/go_pprof_scraper/__about__.py
nsrip-dd: I doubt we'll need to change this much once it's merged, so I'll update this to 1.0.0.

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
FlorentClarret: 7.40 should be ok because I'm not sure 7.41 will be out soon enough

go_pprof_scraper/manifest.json
nsrip-dd: It looks like the CI check doesn't like the new name: https://github.com/DataDog/integrations-extras/actions/runs/3393978080/jobs/5641957063#step:13:18, should I change it back or change it to something else?

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
nsrip-dd: Got it, updated

go_pprof_scraper/manifest.json
FlorentClarret: oops, forgot about that sorry, yes switch it back to `go-pprof-scraper` please

go_pprof_scraper/tests/test_go_pprof_scraper.py
```diff
@@ -0,0 +1,91 @@
...
+    c.check([])
+
+
+def test_check(dd_run_check, aggregator, instance):
+    # type: (Callable[[AgentCheck, bool], None], AggregatorStub, Dict[str, Any]) -> None
+    check = GoPprofScraperCheck("go_pprof_scraper", INIT_CONFIG, [instance])
+    dd_run_check(check)
+
+    aggregator.assert_all_metrics_covered()
+    aggregator.assert_metrics_using_metadata(get_metadata_metrics())
```
FlorentClarret, line 69: You can remove those asserts since there's no metrics

go_pprof_scraper/manifest.json
nsrip-dd: The only consequence of leaving this value as "false" would be that there's not a panel on the integrations page, right? Is it easy to change later? If so, I'm okay leaving it private for now. I've tested this integration myself, but it wouldn't hurt to have some other users internally give it a test run before we widely advertise it.

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
nsrip-dd: I've poked at this a little more, and I'm not sure now where the `datadog_agent.get_config("apm_config")` call is actually getting its config _from_. I tried setting the [`DD_APM_RECEIVER_PORT`](https://github.com/DataDog/datadog-agent/blob/e4e52fff0a85928725e436d6fe092dfa7856b974/pkg/config/config_template.yaml#L1067) env var to something other than the default, and the agent started with that value but `get_config` didn't notice it. Specifically, in [`conftest.py`](https://github.com/DataDog/integrations-extras/pull/1541/files#diff-6321a81caa84fe8dfa36d5a3c1be0d1d118eeaebfd4fda66a530b2334c060da2R27) I set the variable to 9876. Then I added a log line to print the `get_config` value. Then when I start the testing environment I see these two logs:

```
2022-11-06 12:56:26 UTC | TRACE | INFO | (run.go:254 in Infof) | Listening for traces at http://0.0.0.0:9876
...
2022-11-06 12:56:39 UTC | CORE | INFO | (pkg/collector/python/datadog_agent.go:127 in LogMessage) | unknown | (check.py:94) | apm_config: {'enabled': True, 'receiver_port': 8126, 'remote_tagger': True, 'telemetry': {'enabled': True}, 'windows_pipe_buffer_size': 1000000, 'windows_pipe_security_descriptor': 'D:AI(A;;GA;;;WD)'}
```

So the agent is getting configured the way I expect, but the check isn't seeing the correct configuration. The same thing happens if I don't set `DD_APM_ENABLED=true`: the trace agent (where profiles go) won't actually be enabled, but the configuration I read says it is.

nsrip-dd: To be clear, this isn't necessarily a blocker, assuming most people just use the default configuration. But I can't think of any way to work around this in case someone doesn't use the defaults.

go_pprof_scraper/manifest.json
FlorentClarret: Yup! let's keep it as `false` then because yes it's easy to change later

go_pprof_scraper/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Installation
+
+If you are using an Agent version >= 7.21/6.21, follow the instructions below to install the RedisEnterprise check on your host. See the dedicated Agent guide for [installing community integrations][14] to install checks with an [Agent version < v7.21/v6.21][15] or the [Docker Agent][16]:
+
+1. [Download and launch the Datadog Agent][2].
+2. Run the following command to install the integrations wheel with the Agent:
+
+   ```shell
+   datadog-agent integration install -t datadog-redisenterprise==<INTEGRATION_VERSION>
```
FlorentClarret, line 26: 🙈 

```suggestion
   datadog-agent integration install -t datadog-go-pprof-scraper==<INTEGRATION_VERSION>
```

FlorentClarret (changes requested):
The tile is deployed on staging!

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
FlorentClarret: I asked the team and it looks like this is not possible

nsrip-dd: Ok, thanks for looking into it. It's good to have this info, but not critical enough to block this feature. I can investigate further later.

FlorentClarret (approved):
Looks good to me! Thanks for your patience. 

Just waiting for the doc team to have a look and we should be good to go

hestonhoffman (approved):
Small edit, but LGTM

go_pprof_scraper/README.md
```diff
@@ -0,0 +1,79 @@
...
+   ```shell
+   datadog-agent integration install -t datadog-go-pprof-scraper==<INTEGRATION_VERSION>
+   ```
+  You can find the latest version on the [Datadog Integrations Release Page][17]
+
+   **Note**: If necessary, prepend `sudo -u dd-agent` to the install command.
+
+### Configuration
+
+1. Edit the `go_pprof_scraper.d/conf.yaml` file, in the `conf.d/` folder at the root of your Agent's configuration directory to start collecting your go_pprof_scraper performance data. See the [sample go_pprof_scraper.d/conf.yaml][4] for all available configuration options.
```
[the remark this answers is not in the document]
nsrip-dd: Ah, I overlooked this. How about just "Go" performance data? Technically the integration could collect from any program written in Go, as long as the program makes the data available.

FlorentClarret:
Hi @nsrip-dd! May I merge this first version and deploy it?

nsrip-dd:
Yes, thank you!

PR #1680, opened 2022-12-09

ofek:
### Motivation

Part of our tooling revamp


FlorentClarret (changes requested):
you'll need `requests-mock==1.9.2` and disable the e2e tests

PR #1682, opened 2022-12-09

ofek:
### Motivation

Part of our tooling revamp


traefik/hatch.toml
```diff
@@ -0,0 +1,4 @@
+[env.collectors.datadog-checks]
+
+[[envs.default.matrix]]
+python = ["2.7", "3.8"]
```
FlorentClarret, line 4: You'll need to add the two envs for traefik 1.6 and 1.7

FlorentClarret (changes requested):

PR #1683, opened 2022-12-09

ofek:
### Motivation

Part of our tooling revamp


trino/hatch.toml
```diff
@@ -0,0 +1,4 @@
+[env.collectors.datadog-checks]
+
+[[envs.default.matrix]]
+python = ["3.8"]
```
FlorentClarret, line 4: You're missing the trino dep

FlorentClarret (changes requested):

PR #1699, opened 2022-12-15
go_pprof_scraper: add profiler_version tag, fix min_collection_interval default

nsrip-dd:
### What does this PR do?

Adds a missing tag, and fixes a default configuration value:

* Adds a profiler_version tag to profiles uploaded by go_pprof_scraper. The version in the tag is prefixed with "agent-integration" to distinguish Go profiles uploaded by this integration from profiles uploaded by the dd-trace-go client library.
* Sets the default min_configuration_value to 1.

I misunderstood min_collection_interval. From the [docs](https://docs.datadoghq.com/developers/write_agent_check/?tab=agentv6v7#collection-interval):

	Note: If the min_collection_interval is set to 30, it does not mean that
	the metric is collected every 30 seconds, but rather that it could be
	collected as often as every 30 seconds. The collector tries to run the
	check every 30 seconds but the check might need to wait in line,
	depending on how many integrations are enabled on the same Agent. Also
	if the check method takes more than 30 seconds to finish, the Agent
	skips execution until the next interval.

This check is intentionally long-running, and should start again as soon as the previous iteration finishes. I set the default value in the sample config to 1, so the check will be run again as soon as possible. It seems that setting the default to 0, or omitting the value from the config, causes the default to be 15 seconds.

### Motivation

So we can know which versions of the integration are being used, and to fix the defaults so that this runs every minute as intended, rather than ~every 2 minutes (or 1 minute 15 seconds if min_collection_interval is removed) due to the behavior mentioned above.

### Review checklist

- [x] Git history is clean


nsrip-dd:
It seems like the E2E tests might be a bit flaky. [This error](https://dev.azure.com/datadoghq/integrations-extras/_build/results?buildId=122309&view=logs&j=910f82a4-622d-5ace-f877-8b75c045486c&t=12901f7b-620f-5414-43d1-3ae27e5163d7&l=91) says that the check can't hit the agent's profiling endpoint. It's odd because the agent which is serving that endpoint is the same one running the check... And the Python 2 test immediately before it worked. I haven't seen this problem at all locally.

go_pprof_scraper/datadog_checks/go_pprof_scraper/data/conf.yaml.example
```diff
@@ -69,13 +69,17 @@ instances:
     #
     service: <SERVICE>
 
-    ## @param min_collection_interval - number - optional - default: 60
-    ## How frequently profiles are collected.
-    ##
+    ## @param min_collection_interval - number - optional - default: 1
```
nsrip-dd, line 72: Despite what it says here, the default if I leave out this parameter is 15 seconds. I'm guessing it's set somewhere else before the check is started? I'd like the default to actually be 1, but as long as these docs are clear and the sample value below is 1, that's okay with me.

go_pprof_scraper/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG - Go-pprof-scraper
 
+## 1.0.2 / 2022-12-16
+
+* [Fixed] Default `min_collection_interval` value is decreased so there isn't a
+  long pause between collecting profiles.
```
yzhan289, line 6:
```suggestion
* [Fixed] Decrease `min_collection_interval` value to prevent long pause between collecting profiles.
```

go_pprof_scraper/datadog_checks/go_pprof_scraper/data/conf.yaml.example
yzhan289: Yup by default the `min_collection_interval` is [15 seconds](https://docs.datadoghq.com/developers/write_agent_check/?tab=agentv6v7#collection-interval). Just note that having the value be 1 will not _guarantee_ every second:

> If the min_collection_interval is set to 30, it does not mean that the metric is collected every 30 seconds, but rather that it could be collected as often as every 30 seconds. The collector tries to run the check every 30 seconds but the check might need to wait in line, depending on how many integrations are enabled on the same Agent. Also if the check method takes more than 30 seconds to finish, the Agent skips execution until the next interval.

nsrip-dd: Got it, thanks! This integration runs for a long time (60 seconds by default), so my intention is just that once it's done, it starts again as soon as possible. If it's delayed by a second or two, that's okay. Even 15 seconds is okay with me, as long as the user knows it's possible to make it happen more frequently :)

nsrip-dd:
I noticed from [these docs](https://github.com/DataDog/integrations-core/blob/master/docs/developer/process/integration-release.md#creating-the-release) and some [previous](https://github.com/DataDog/integrations-extras/pull/1593) [examples](https://github.com/DataDog/integrations-extras/pull/1592) that integrations usually seem to get released in separate pull requests. Should I remove the `__about__.py` and `CHANGELOG.md` updates from this PR?

FlorentClarret:
> I noticed from [these docs](https://github.com/DataDog/integrations-core/blob/master/docs/developer/process/integration-release.md#creating-the-release) and some [previous](https://github.com/DataDog/integrations-extras/pull/1593) [examples](https://github.com/DataDog/integrations-extras/pull/1592) that integrations usually seem to get released in separate pull requests. Should I remove the `__about__.py` and `CHANGELOG.md` updates from this PR?

You can but this is not a hard requirements for extras, only for -core.

nsrip-dd:
I don't have permission to click the merge button, can somebody please merge if this is good to go?

nsrip-dd:
Apologies, I seem to have screwed up something when trying to pull in more recent changes from master. Marking this WIP again until I fix it.

nsrip-dd:
Okay, I think I fixed it? I updated the changelog to include the recently merged fix from #1703

fanny-jiang (approved):

PR #1743, opened 2023-01-27
[Lighthouse] Fix bug with metric tags when specifying both `tags` and `urls` in config

beepboop271:
#### Issue

Fixes a bug where if multiple URLs were given in `urls`, and common tags to apply to all metrics were given in `tags`, metrics for each url would have tags from every url before it in the list too.

For example:

Screen Shot 2023-01-30 at 1 54 37 PM

Each instance's URLs has an growing list of URL tags, like from\
`url:https://www.super.com`, then\
`url:https://www.livesuper.com/shop,url:https://www.super.com`, then\
`url:https://www.livesuper.com/home,url:https://www.livesuper.com/shop,url:https://www.super.com`, and so on rather than just being tagged with the one URL it's from.

As another example, the test added in this PR, `test_check_urls_tags`, fails on the old version of `lighthouse.py` with the following:

```
E   AssertionError: Needed at least 1 candidates for 'lighthouse.accessibility', got 0
E     Expected:
E             MetricStub(name='lighthouse.accessibility', type=None, value=92, tags=['key:value', 'name:test', 'url:https://www.datadoghq.com'], hostname=None, device=None, flush_first_value=None)
E     Similar submitted:
E     Score   Most similar
E     0.98    MetricStub(name='lighthouse.accessibility', type=0, value=92.0, tags=['key:value', 'name:test', 'url:https://www.google.com'], hostname='', device=None, flush_first_value=False)
E     0.95    MetricStub(name='lighthouse.accessibility', type=0, value=92.0, tags=['key:value', 'name:test', 'name:test', 'url:https://www.datadoghq.com', 'url:https://www.google.com'], hostname='', device=None, flush_first_value=False)
(rest omitted)
```

Note the metric with `tags=['key:value', 'name:test', 'name:test', 'url:https://www.datadoghq.com', 'url:https://www.google.com']`, which contains the tags for the datadog URL (expected), but also the tags for the google URL before it (bad).

Metrics should only be tagged with the URL they came from.

#### Fix

This is because the tags (in `lighthouse.py`) are constructed from `instance.get("tags", [])` within each iteration of the loop through the urls. If `instance` does have `tags` set, it would `.get` the same list reference in each iteration, and the tags are appended to the same list. So each iteration 'pollutes' the tags of the remaining iterations.

Fix: create a new tags list each iteration since the urls and thus tags are always different.

### Motivation

Metrics are being tagged incorrectly.

### Review checklist

- [ ] Git history is clean


lighthouse/tests/test_lighthouse.py
```diff
@@ -58,6 +58,47 @@ def test_check(aggregator, instance):
...
+        "urls": ["https://www.google.com", "https://www.datadoghq.com"],
+        "name": "test",
+        "tags": [common_tag],
+    }
+    lighthouse_check = LighthouseCheck("lighthouse", {}, {})
+    LighthouseCheck._get_lighthouse_report = MagicMock(side_effect=mock_get_lighthouse_report)
+
+    lighthouse_check.check(instance)
+
+    expected_tags = [[common_tag, "url:" + url, "name:" + instance["name"]] for url in instance["urls"]]
```
yzhan289, line 74: To make sure I understand the expected result, you want to see `tags=['key:value', 'name:test', 'name:test', 'url:https://www.datadoghq.com', 'url:https://www.google.com']` but not `tags=['key:value', 'name:test', 'name:test', 'url:https://www.datadoghq.com,url:https://www.google.com']`?

yzhan289 (commented):
Hey 👋 , sorry for the delay. Left some comments!

lighthouse/tests/test_lighthouse.py
beepboop271: The expected result is `tags=['key:value', 'name:test', 'url:https://www.datadoghq.com']` for metrics created when it runs the check on the DD url, and `tags=['key:value', 'name:test', 'url:https://www.google.com']` for metrics created when it runs the check on the Google url.

Hopefully the commit I pushed makes the expectation clearer.

---

The test failure I pasted in the PR description is what happens when I run this new test on the previous version of the check code, and demonstrates the undesired previous behaviour in the "similar submitted" metrics.

It created `tags=['key:value', 'name:test', 'url:https://www.google.com']` on the Google URL (the first in the list), but the DD URL (the second in the list) had metrics tagged with `tags=['key:value', 'name:test', 'name:test', 'url:https://www.datadoghq.com', 'url:https://www.google.com']`, which contains undesired tags from the previous URL in the list, the Google one.

aianus:
Hi @yzhan289 any update on this PR?

PR #1753, opened 2023-02-08
changed dashboard to work with last pipeline

raulpopa:
Dashboard update. Previous dashboard used usr.name and new pipeline does not remap it, so we're using typingdna.username instead, also, the new dashboard uses the high density mode.

### What does this PR do?

Only changes the dashboard to work with new pipeline remaps.


alexvetras (approved):
Approved

moreauowen (approved):

PR #1796, opened 2023-03-29
Fix Pihole Integration to support Pihole v5.x and Datadog Agent v7+

troyfontaine:
### What does this PR do?

This fixes the Pihole integration to support both Datadog Agent v7 and Pihole v5 (as the API requires an API token currently to be able to query metrics.

### Motivation

I was attempting to use this integration on my homelab and ran into a number of errors, starting from the attempt to query the API (bad URL), to the authentication required now and finished with the necessity to pass the metrics as floating point numbers instead of the string they originally were configured to use.

### Review checklist

- [ ] Git history is clean

### Additional Notes


pihole/README.md
```diff
@@ -22,9 +22,10 @@ For Agent v7.21+ / v6.21+, follow the instructions below to install the Pi-hole
 
 ### Configuration
 
-1. Edit the `pihole.d/conf.yaml` file, in the `conf.d/` folder at the root of your Agent's configuration directory to start collecting your Pi-hole performance data. See the [sample pihole.d/conf.yaml][7] for all available configuration options.
+1. In the root of your Agent's configuration directory copy the `conf.d/pihole.d/conf.yaml.example` to `conf.d/pihole.d/conf.yaml`
+2. Edit the `conf.d/pihole.d/conf.yaml` by adding an instance and including both the host and API token in order to start collecting your Pi-hole metrics. See the [sample pihole.d/conf.yaml][7] for all available configuration options.
```
hestonhoffman, line 26:
```suggestion
1. In the root of your Agent's configuration directory, copy `conf.d/pihole.d/conf.yaml.example` to `conf.d/pihole.d/conf.yaml`.
2. Edit `conf.d/pihole.d/conf.yaml` by adding an instance and including both the host and API token in order to start collecting your Pi-hole metrics. See the [sample pihole.d/conf.yaml][7] for all available configuration options.
```

hestonhoffman (approved):
A couple of small docs edits, but this is good to go from the docs perspective. Thanks for the contribution!

troyfontaine:
> A couple of small docs edits, but this is good to go from the docs perspective. Thanks for the contribution!

Thanks!

troyfontaine:
I'm having some difficulty with fixing the tests-I've added in some logic to handle the difference between the numbers in the test and those actually being seen in API response from my PiHole instance-but getting the good_response to match up isn't working so well as I'm not very familiar with the test library and how to get the mock up to match up to the expected response properly.

monganai:
👋  Hey @troyfontaine

Thanks for your contributions here!

On my side, I have the integration running with Agent v7.4x.x  and i'm not using a token (I don't think it required one on initial setup)  In this case, the `token` should be optional rather than required for a given instance of the check.

In terms of testing the integration, the existing test instances target the output of hitting the different endpoints in [this repo](https://github.com/monganai/fake-hole/blob/master/fake-hole.py).   The [test suite](https://github.com/DataDog/integrations-extras/tree/master/pihole/tests) currently spins up the `fake-hole`  Docker container to test against, however I don't think it will handle being passed a token at this time.

Perhaps adding a new route in / updating `fake-hole.py`, to handle the addition of `?summary&auth=token` would be the most straight forward solution?  What are your thoughts?

troyfontaine:
> 👋 Hey @troyfontaine
> 
> Thanks for your contributions here!
> 
> On my side, I have the integration running with Agent v7.4x.x and i'm not using a token (I don't think it required one on initial setup) In this case, the `token` should be optional rather than required for a given instance of the check.
> 
> In terms of testing the integration, the existing test instances target the output of hitting the different endpoints in [this repo](https://github.com/monganai/fake-hole/blob/master/fake-hole.py). The [test suite](https://github.com/DataDog/integrations-extras/tree/master/pihole/tests) currently spins up the `fake-hole` Docker container to test against, however I don't think it will handle being passed a token at this time.
> 
> Perhaps adding a new route in / updating `fake-hole.py`, to handle the addition of `?summary&auth=token` would be the most straight forward solution? What are your thoughts?

I'll get on it!  Thanks!

troyfontaine:
@monganai I've made the token completely optional-if it is detected, it will use it, otherwise the integration will just keep trucking along.  I'll try and find some time next week to add the token route to the test container.  It looks like because I use a password on my pihole instances, the [token is required](https://pi-hole.net/blog/2022/11/17/upcoming-changes-authentication-for-more-api-endpoints-required/#page-content).  I can include this information in the readme as part of this PR or as a new PR if preferred.  Cheers!

pihole/CHANGELOG.md
```diff
@@ -1,2 +1,18 @@
 # CHANGELOG - pihole
 
+## [3.14.2] - 2023-03-29
```
iliakur, line 3:
```suggestion
## 3.14.2 - 2023-03-29
```

pihole/CHANGELOG.md
```diff
@@ -1,2 +1,18 @@
...
+### Added
+
+- Support for Pihole API Token authentication
+- Support for Pihole version 5.0+
+
+### Fixed
+
+- Convert string numbers captured from Pihole to floating point as required by the Datadog Agent
+
+## [3.14.1] - Original Release
```
iliakur, line 14:
```suggestion
## 3.14.1 - Original Release
```

iliakur (changes requested):
Note that in order to release this you also need to bump the version number in `__about__.py`.

PR #1808, opened 2023-04-10
Don't allow errors to make the entire check fail

ofek:
### Additional Notes

Also added logging


ofek:
https://github.com/DataDog/integrations-extras/pull/1809

PR #1933, opened 2023-06-26
Update formatting for changelogs

swang392:
### Motivation

Make changelogs more readable

### DIFF
--- a/filebeat/CHANGELOG.md
+++ b/filebeat/CHANGELOG.md
@@ -1,37 +1,59 @@
 # CHANGELOG - Filebeat Integration
 
-1.3.0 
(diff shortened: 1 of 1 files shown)


filebeat/CHANGELOG.md
```diff
@@ -1,37 +1,51 @@
...
+
+
+## 1.1.0 
+***Added***: 
+
+* Add option to ignore the registry file check
+
+
+## 1.0.0 
+
```
nmuesch, line 22: Should we try to find something to add here so that its not an empty version?

nmuesch (approved):

PR #2236, opened 2023-12-18
Fix out of sync files

steveny91:
### What does this PR do?
Fix some out of sync files that were flagged by CI


Kyle-Neale (approved):

iliakur (commented):
@steveny91 should this get a changelog entry?

steveny91:
> @steveny91 should this get a changelog entry?

🤔 I personally would have skipped the change log since it's just resyncing the models that should have been done at a previous commit (when we update pydantic) but I can be convinced otherwise.

PR #2261, opened 2024-01-19
Apply new linter change (ordering facet attributes)

thibaultkrebs:
# Context

We are working on a feature to sort facet attributes (https://datadoghq.atlassian.net/browse/LOGSC-746).

This PR is to apply change with the new linter check.


FlorentClarret (approved):

nmuesch (approved):
Should be no functional changes here and is only a sorting change

PR #2368, opened 2024-04-30
adds support to query baseline from backend before using it in queries

bhalder:
### What does this PR do?

- In the upcoming version of Fiddler, the default baseline name were updated.
- This PR adds a query to get baseline names from backend, before using it in the queries

### Motivation

- To ensure continued support of existing DataDog customers using Fiddler 

Screen Shot 2024-05-01 at 12 16 38 PM
Screen Shot 2024-05-01 at 12 22 05 PM

### Review checklist

- [x] Git history is clean
- [x] If PR impacts documentation, docs team has been notified or an issue has been opened on the [documentation repo](https://github.com/DataDog/documentation/issues/new)
- [x] If this PR includes a log pipeline, please add a description describing the remappers and processors.


fiddler/datadog_checks/fiddler/check.py
```diff
@@ -265,6 +283,8 @@ def _run_queries(self, project: str, model: str) -> (Dict[str, Any], list):
         start_time = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(start_time_epoch / 1000))
         end_time = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(end_time_epoch / 1000))
 
+        baseline_name = self._get_baseline_name(project, model)
```
gnanesh-fiddler, line 286: I think it makes more sense to call it only when baseline name is required for the metric. We should move it inside the if condition  - https://github.com/DataDog/integrations-extras/pull/2368/files#diff-c10da9b700759108e912edfd1bab6da1ba032c6679973175e6f8ea898da57ce6R321.

bhalder: Actually, this one is better, because we only get baseline name for a model once instead of for each metric.
Moving this under the if condition will mean multiple calls to get baseline name.

Issue #2529, opened 2024-11-04
Change prefix for `apollo` integration


BrynCooke:
I work for Apollo and am following up on an issue raised by a user regarding tag management.

The Apollo integration specifies that metrics that have a prefix `apollo.` are part of an integration:
https://github.com/DataDog/integrations-extras/blob/master/apollo/manifest.json#L35

Apparently having this prefix reserved limits users' from managing tags via the DD UI.

> After further investigation, we’ve identified that the reason you’re unable to manage tags for these custom metrics is that they are using the apollo. prefix. This prefix is reserved for Apollo out-of-the-box (OOTB) metrics, which is why tag management is restricted.

Some of our products emit metrics that use the `apollo.` prefix, and users should be able to manage tags.
Are there any avenues for us to change the integrated metrics to the following prefixes:
* `apollo.operations.`
* `apollo.engine.`

What would be needed from our side to enable this to happen?

abernix:
I opened a draft PR to try and address this, but do still need guidance.

PR #2539, opened 2024-11-19
[Apollo] Tighten namespace around metrics

abernix:
### What does this PR do?

This aims to tighten the namespace which are part of our DataDog Marketplace integration to **just** the set of metrics which are emitted (from the SaaS side of our Apollo product) which pertain to the Marketplace integration rather than those that are emitted by our (different) on-prem product, Apollo Router which emits its metrics into an `apollo.router.`-prefixed namespace.  This `apollo.router.` namespace should _not_ prevent users from losing control over the management of tags in their UI.

What this PR does not do is remove `apollo.engine.` from the list of `checks`, though it is tempting to do that at the same time.  What is not clear is what the implications of that are as this is a deprecated namespace.

Relates to https://github.com/DataDog/integrations-extras/issues/2529

### Motivation

We're opening this only because we're finding customers who are using our on-prem metrics are being forbidden by the DataDog UI to adjust those metrics in the way they would manage tags on other metrics.

### Review checklist

- [ ] Git history is clean

### Additional Notes

Happy to help satisfy the Review checklist, but primarily looking for feedback if this will accomplish what we want, or if there is another way this should be done.


apollo/manifest.json
```diff
@@ -33,7 +33,7 @@
         "creates_events": false
       },
       "metrics": {
-        "prefix": "apollo.",
+        "prefix": "apollo.operations.",
```
abernix, line 36: I'm curious if the manifest supports having an array here?  It would seem safer conceptually to me — to preserve behavior of our legacy `apollo.engine.` namespace — to have this be `["apollo.operations.", "apollo.engine."`]` and _not_ add `apollo.router.` to that list (which is the primary set we're trying to unblock tag-editing access for).

abernix:
I'll switch this to a draft, but we could really use guidance on what the most practical path forward would be to "preserve whatever is already working on `apollo.engine.` for folks who won't migrate off that for a while, but also unblocks the ability to edit `apollo.router.` tag metrics.  We're of course happy if `apollo.operations.*` continues to have that tag-editing limitation, but just don't know how to craft that into the manifest's desired format.)

PR #2548, opened 2024-11-25
Add missing permissions in GitHub Actions workflows.

nubtron:
### What does this PR do?

This PR updates master.yml and publish-test-results-pr.yml to include the missing permissions as we transition read-only permissions by default. 

In particular, publish-test-results-pr.yml requires explicit read permissions, since defining permissions in the "permissions" section sets all other permissions to none.


steveny91 (approved):

nubtron:
@steveny91 Thanks for the review!

PR #2570, opened 2024-12-19
Fix logstash metrics descriptions

schew2381:
I noticed some of the logstash metric descriptions are mismatched


brett0000FF (approved):
Good catch! Thanks for the contribution. 🚀

PR #2653, opened 2025-04-11
Nextcloud: make the `nextcloud.system.apps.*` metrics optional

VJean:
### What does this PR do?

Beginning with Nextcloud 28, the monitoring endpoint (the `serverinfo` app) [no longer provides information about available app updates](https://docs.nextcloud.com/server/latest/admin_manual/release_notes/upgrade_to_28.html#monitoring), as gathering the data always involves at least one external request to `apps.nextcloud.com`.

The test image was pinned to an earlier, outdated version. However, when trying to set up the Nextcloud integration against a recent server (version 31 for instance), the check would fail as it is unable to find and parse the `nextcloud.system.apps.*` metrics from the stats call output.

This PR:
* adds an option, `apps_stats`, to control whether to fetch these additional stats. Defaults to false, to match Nextcloud's current behavior
  * adds a spec file, `nextcloud/assets/configuration/spec.yaml` to describe the check's configuration
  * adds the generated matching default example conf, `nextcloud/datadog_checks/nextcloud/data/conf.yaml.example`
* adds a test
* improves the test suite reliability
  * unpins the tested Nextcloud image to ensure the check continues to be compatible with newer versions
  * wait for the service to be available first before trying to run commands in the container

### Motivation

The check fails for the latest supported Nextcloud servers.

### Review checklist

- [x] Git history is clean

### Additional Notes

Not sure if I should update the Changelog and version now or if it is done as part of a separate release process.


estherk15 (approved):

nextcloud/CHANGELOG.md
```diff
@@ -1,2 +1,14 @@
 # CHANGELOG - Nextcloud
 
+# 1.1.0 / 2025-04-12
```
steveny91, line 3: Would you mind leaving this change log as a small 1 sentence change? Can you move the section explaining the change to the readme instead? I think it should be customer facing.

VJean: @steveny91 Thanks! I shortened the changelog entry and added a note in the Readme file. Let me know if that's better now

nextcloud/CHANGELOG.md
```diff
@@ -1,2 +1,8 @@
 # CHANGELOG - Nextcloud
 
+# 1.1.0 / 2025-04-25
```
steveny91, line 3:
```suggestion
# 2.0.0 / 2025-04-25
```

I think a changed type changelog entry warrants a major bump. I think that would be better suited in this case since with the same config we no longer can get the same amount of metrics. But I guess this would only be for people on newer version of the tech? So a user with an old version of next cloud with the new version of the check with apps_stats disabled can still get the apps.* metrics? I guess this change is only aimed at newer version users? Trying to see if this really warrants a changed type entry or can it be an added.

VJean: So users on Nextcloud < 28 will indeed not get the `apps.*` metrics anymore, unless they enable `apps_stats`. With the same config, they can no longer get the same amount of metrics.

I'm ok with a major version bump in this regard.

Adding a note that [Nextcloud 27 was EOL in 2024-06 and Nextcloud 28 itself has been EOL since 2024-12](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule#eol-versions) for what it's worth.

steveny91 (approved):

PR #2696, opened 2025-05-20
[ECOINT-185] Fix Modal input metrics tags

xpbowler:
Fixes https://github.com/DataDog/integrations-extras/pull/2689 by removing the `timestamp` tag and adding `task_id` tag.

The estimated max cardinality per metric integration is **1000**.
(`modal.input_events.elapsed_time_us`, `modal.input_events.total_inputs`, `modal.input_events.successes` metrics)


modal/metadata.csv
dd-dominic, line 1: How will the task_id be valuable for customers? Should this change be captured on the dashboard?

dd-dominic (changes requested):

modal/metadata.csv
xpbowler: In our system, a task is a 1-on-1 mapping with a user's containers - to which they are limited at 1000 containers. It's useful for customers because they might want to filter by task_id to figure out if a particular container is running slower than expected. The change shouldn't be captured on the dashboard because a user won't filter by this tag by default.

xpbowler: @dd-dominic updated `task_id` to be `container_id` for more clarity for customers.

dd-dominic (approved):

PR #2777, opened 2025-08-27
[SAASINT-4526] DDS: Zscaler Internet Access Integration v2.0.0

vrajpatel-crest:
### What does this PR do?

- This update extends the integration to include following additional log types from ZIA:
  - Admin Audit Logs
  - SaaS Security Logs
  - SaaS Security Activity Logs
  - Endpoint DLP Logs
  - Email DLP Logs
  - Alerts
- Support for additional fields in already supported log types
- Additional OOTB Dashboards
- OOTB Log Detection rules
- Updated Pipeline

### Motivation

- This integration is webhook based so Crawler will not be used.
- The Datadog team will implement the appropriate component for generating URLs. A URL will be created having the ddsource query parameter set to zscaler.

### Review checklist

- [ ] Git history is clean

### Additional Notes

### DIFF
--- a/zscaler/assets/logs/zscaler.yaml
+++ b/zscaler/assets/logs/zscaler.yaml
@@ -597,7 +597,7 @@ pipeline:
           target: zscaler.url_full
           replaceMissing: false
         - type: attribute-remapper
-          name: Map `zscaler.url_full` to `http.url`
+          name: Map `url_full` to `http.url`
           enabled: true
           sources:
             - zscaler.url_full
@@ -614,7 +614,7 @@ pipeline:
           target: http.url_details
           normalizeEndingSlashes: false
         - type: grok-parser
-          name: Convert `threatseverity` field into a lowercase
+          name: Convert `threatseverity` field into lowercase
           enabled: true
           source: zscaler.threatseverity
           samples:
@@ -710,7 +710,7 @@ pipeline:
           preserveSource: true
           overrideOnConflict: false
         - type: grok-parser
-          name: Convert `threatseverity` field into a lowercase
+          name: Convert `threatseverity` field into lowercase
           enabled: true
           source: zscaler.threatseverity
           samples:
@@ -868,8 +868,7 @@ pipeline:
           preserveSource: false
           overrideOnConflict: false
         - type: string-builder-processor
-          name: Reset `postaction.cloudNssSiemConfiguration.connectionHeaders`
-            attribute
+          name: Reset `postaction.cloudNssSiemConfiguration.connectionHeaders` attribute
           enabled: true
           template: "%{resolve_to_empty_string}"
           target: zscaler.postaction.cloudNssSiemConfiguration.connectionHeaders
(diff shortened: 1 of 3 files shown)


zscaler/CHANGELOG.md
```diff
@@ -1,2 +1,19 @@
 # CHANGELOG - Zscaler
 
+## 1.0.0
+
+**Added**
+
+* Initial Release
+
+## 2.0.0 / 2025-08-27
```
tkell, line 3: Please move this to the top, so we have the most recent changes first!

zscaler/manifest.json
```diff
@@ -32,19 +76,24 @@
       "events": {
         "creates_events": false
       },
-      "metrics": {
-        "prefix": "zscaler.",
-        "check": [],
-        "metadata_path": "metadata.csv"
-      },
```
tkell, line 39: I'm sure this is correct, but want to make sure that we're removing metrics — it looks like we never had any?

vrajpatel-crest: Yes, there are no metrics.

zscaler/assets/logs/zscaler.yaml
```diff
@@ -299,205 +402,825 @@ pipeline:
...
+          samples:
+            - "Alert ID: 567890 | Alert Status: Ended |  Rule Name: Self Created
+              | Event Type: Malicious URLs | Indexed By:
+              www.institutolockmann.com.br | ZIA-URL:
+              https://admin.zscalertwo.net/#alert-details?alertId=567890&startT\
+              ime=1755525600"
+          grok:
+            supportRules: ""
+            matchRules: 'indexed_by_rule %{regex(".*(?= Indexed By)")} Indexed By:
+              %{notSpace:indexed} \|.*'
```
tkell: OK, the grok parsers that pull from the Description should be tweaked.  We need to use non-greedy matchers as much as possible, and not use `data` as much as we can.

For example, we've got `indexed_by_rule %{regex(".*(?= Indexed By)")} Indexed By: %{notSpace:indexed}` — we should do `indexed_by_rule .+?Indexed By: %{notSpace:indexed} \|.*` instead, so we're not using a greedy matcher.

Likewise, for `Event Type`, we should do `event_type_rue .+?Event Type: %{regex("[A-z ]*?"): event_type} \|.*` — note that we've replaced `data:` with a more explicit regex, as `data:` can be too general a matcher.

vrajpatel-crest: We have updated the GROK pattern to match up to the word, but to match the value, we need to use the data since the value contains special characters.

tkell: Great, thanks!

We don't need the extra `\` though — you can try it yourself:

Screenshot 2025-09-02 at 08 49 42

vrajpatel-crest: We used a single backslash `\`, but during pipeline export, YAML automatically added an extra backslash and wrapped the value in double quotes.

image

tkell: Ah, got it — and I see in the test that it is working, looks good!

tkell (approved):

PR #2801, opened 2025-09-20
Fix resilience4j e2e service check test

willianccs:
### What does this PR do?

- Remove incorrect pytest.raises(Exception) wrapper that was causing test failures
- Update test to use dd_run_check instead of dd_agent_check for proper fixture usage
- Add mock_prometheus_metrics fixture to avoid real network connections during testing
- Bump version to 1.1.1

The test now properly validates service check emission without attempting to connect to real endpoints or expecting erroneous exceptions.

### Motivation

Create a new release with latest changes in metrics.

### Review checklist

- [ ] Git history is clean

### Additional Notes


willianccs:
@steveny91 could you help here?

This is a fix related to the previous [PR](https://github.com/DataDog/integrations-extras/pull/2771) that didn't create the `1.1.0` release version :disappointed: 

This fix addresses the e2e test issues that were preventing the integration from being properly released, and includes the necessary version bump and compatibility adjustments.

Thank you in advance!

resilience4j/hatch.toml
```diff
@@ -1,4 +1,4 @@
 [env.collectors.datadog-checks]
 
 [[envs.default.matrix]]
-python = ["3.12"]
+python = ["3.11"]
```
NouemanKHAL: The datadog-agent is shipped with python 3.12, why this change?

resilience4j/tests/test_e2e.py
```diff
@@ -9,9 +9,8 @@
 pytestmark = [pytest.mark.usefixtures("dd_environment"), pytest.mark.e2e]
 
 
-def test_e2e_resilience4j_check_e2e_assert_service_check(dd_agent_check, aggregator, instance):
-    with pytest.raises(Exception):
-        dd_agent_check(instance, rate=True)
+def test_e2e_resilience4j_check_e2e_assert_service_check(dd_run_check, aggregator, check, mock_prometheus_metrics):
```
NouemanKHAL: This test needs to be rewritten to assert for one scenario at a time, when we expect a service check to be OK, and when we don't expect any service check, the current check will always pass or be skipped, which is adding any value in my opinion.

resilience4j/hatch.toml
willianccs: You're absolutely right! I've removed the Python 3.11 workaround and implemented proper fixes

resilience4j/tests/test_e2e.py
willianccs: After investigating further, I discovered when connection fails, OpenMetrics checks raise exceptions instead of sending CRITICAL service checks
**Fix implemented:** Updated `test_e2e_resilience4j_service_check_critical_on_connection_error` to expect exceptions using `pytest.raises(Exception)`
This now reflects the actual behavior of the resilience4j integration which is OpenMetrics-based.

resilience4j/pyproject.toml
```diff
@@ -26,10 +26,11 @@ classifiers = [
     "License :: OSI Approved :: BSD License",
     "Private :: Do Not Upload",
     "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
     "Topic :: System :: Monitoring",
 ]
 dependencies = [
-    "datadog-checks-base>=32.6.0",
+    "datadog-checks-base>=37.0.0",
```
NouemanKHAL, line 33: question: any reason behind this bump? If so, there should be a changelog reflecting it.

willianccs: The bump fixes a CI pipeline failure with Python 3.12. The `--compat` flag forces minimum version `32.6.0` which has `ddtrace==1.11.2` that doesn't compile on Python 3.12 (Cython errors: PyInt_Check undefined).

Version 37.0.0 uses Python 3.12 compatible dependencies. This is a technical fix for CI compatibility, not a user-facing change, but I can add a changelog entry if preferred.

willianccs: @NouemanKHAL do I have to update the changelog?

nubtron: @willianccs yes, an changelog entry is needed in this case because we're updating a dependency. 

Integrations are shipped as Python wheels, and we want to avoid wheel updates without changelog entries. As a rule of thumb: tests, assets, or documentation changes don’t need changelogs, but any non-test Python code changes or dependency updates do.

nubtron (changes requested):
Looks good to me other than the missing changelog entry!

PR #2820, opened 2025-10-07
Fix `go_pprof_scraper` Unix socket fallback when default socket path doesn't exist

Kyle-Neale:
### What does this PR do?
This PR fixes the `go_pprof_scraper` integration to gracefully handle the case where the Datadog Agent has a default Unix socket path configured (`apm_config.receiver_socket`) but the socket file doesn't actually exist. The integration now:
- Validates socket existence before attempting to use Unix socket communication
- Falls back to TCP automatically when the socket doesn't exist or connection fails
- Caches the working upload method to avoid repeated failed connection attempts on subsequent check runs

### Motivation
Recent changes to the Datadog Agent changed `apm_config.receiver_socket` from returning an empty value when not set to returning a default path (e.g., `/var/run/datadog/apm.socket`). This caused the integration to attempt Unix socket connections even when the Agent is running in TCP mode, resulting in:
```
FileNotFoundError: [Errno 2] No such file or directory
```

### Review checklist

- [ ] Git history is clean

### Additional Notes

### DIFF
--- a/go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
+++ b/go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
@@ -209,6 +209,7 @@ class GoPprofScraperCheck(AgentCheck):
                     self._preferred_upload_method = "unix"
                     self.log.debug("Successfully uploaded profiles via Unix socket")
                     # Success - no need to try TCP
+                    self.service_check("can_connect", AgentCheck.OK, tags=[])
                     return
                 except (ConnectionError, FileNotFoundError) as e:
                     # Remember that Unix socket doesn't work, use TCP from now on
@@ -239,6 +240,8 @@ class GoPprofScraperCheck(AgentCheck):
                 # Cache TCP as the working method
                 self._preferred_upload_method = "tcp"
                 self.log.debug("Successfully uploaded profiles via TCP")
+                self.service_check("can_connect", AgentCheck.OK, tags=[])
+                return
             except Exception as e:
                 # If both Unix socket and TCP failed, raise the most recent error
                 if last_error:
@@ -271,8 +274,3 @@ class GoPprofScraperCheck(AgentCheck):
                 message="Request failed: {}, {}".format(self.url, e),
             )
             raise
-
-        # If your check ran successfully, you can send the status.
-        # More info at
-        # https://datadoghq.dev/integrations-core/base/api/#datadog_checks.base.checks.base.AgentCheck.service_check
-        self.service_check("can_connect", AgentCheck.OK, tags=[])
--- a/go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
+++ b/go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
@@ -189,13 +189,10 @@ class GoPprofScraperCheck(AgentCheck):
             # TODO: container ID? If we can get it, it should be added as a
             # "Datadog-Container-ID" header to the request.
 
-            # Use the preferred upload method if we've already determined it
-            upload_successful = False
-            last_error = None
-
             # Try Unix socket first if:
             # 1. It's configured (socket exists)
             # 2. We haven't determined a preference yet OR we know Unix socket works
+            last_error = None
             should_try_unix = self.trace_agent_socket and (
                 self._preferred_upload_method is None or self._preferred_upload_method == "unix"
             )
@@ -209,9 +206,10 @@ class GoPprofScraperCheck(AgentCheck):
                     session = requests_unixsocket.Session()
                     r = session.post(self.trace_agent_socket, files=files)
                     r.raise_for_status()
-                    upload_successful = True
                     self._preferred_upload_method = "unix"
                     self.log.debug("Successfully uploaded profiles via Unix socket")
+                    # Success - no need to try TCP
+                    return
                 except (ConnectionError, FileNotFoundError) as e:
                     # Remember that Unix socket doesn't work, use TCP from now on
                     self._preferred_upload_method = "tcp"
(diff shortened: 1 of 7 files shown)


nsrip-dd:
For some reason, when I run the end-to-end test locally, I see this error:

```
% ddev env start --dev --agent datadog/agent:7 go_pprof_scraper py3.12
% ddev env agent go_pprof_scraper py3.12 check
=== Service Checks ===
[
  {
    "check": "go_pprof_scraper.can_connect",
    "host_name": "comp-yq6wk6w9dv",
    "timestamp": 1759842845,
    "status": 2,
    "message": "the trace agent is not enabled",
    "tags": [
      "service:testing"
    ]
  }
]
```

I see that in CI as well, and I even get this running with the commit prior to this change 🤔 Are you seeing this locally as well? It's been quite a bit since I worked on this so I'm not totally sure what to try.

Kyle-Neale:
> For some reason, when I run the end-to-end test locally, I see this error:
> 
> ```
> % ddev env start --dev --agent datadog/agent:7 go_pprof_scraper py3.12
> % ddev env agent go_pprof_scraper py3.12 check
> === Service Checks ===
> [
>   {
>     "check": "go_pprof_scraper.can_connect",
>     "host_name": "comp-yq6wk6w9dv",
>     "timestamp": 1759842845,
>     "status": 2,
>     "message": "the trace agent is not enabled",
>     "tags": [
>       "service:testing"
>     ]
>   }
> ]
> ```
> 
> I see that in CI as well, and I even get this running with the commit prior to this change 🤔 Are you seeing this locally as well? It's been quite a bit since I worked on this so I'm not totally sure what to try.

Hmm interesting - I couldn't get it to work locally in my initial attempts and it fails on master with that same error. I know `ddev`, when starting the docker container, overwrites changes made to APM configuration so the trace Agent is disabled by default. I made some changes to my `ddev` to fix this issue and it passes locally for me - I was going to push the changes up later:
```
cmd [1] | pytest -vv --benchmark-skip --tb short -m e2e
============================= test session starts ==============================
platform darwin -- Python 3.12.8, pytest-8.1.1, pluggy-1.6.0 -- /Users/kyle.neale/Library/Application Support/hatch/env/virtual/datadog-go-pprof-scraper/lkxMrBxI/py3.12/bin/python
cachedir: .pytest_cache
benchmark: 5.1.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/kyle.neale/go/src/github.com/DataDog/integrations-extras/go_pprof_scraper
configfile: pyproject.toml
plugins: asyncio-0.23.8, memray-1.8.0, datadog-checks-dev-35.2.0, ddtrace-3.12.5, mock-3.15.1, flaky-3.8.1, benchmark-5.1.0, cov-7.0.0
asyncio: mode=Mode.STRICT
collecting ... collected 7 items / 6 deselected / 1 selected

tests/test_go_pprof_scraper.py::test_e2e PASSED                          [100%]

======================= 1 passed, 6 deselected in 2.84s ========================
```

nsrip-dd:
The check works for me locally. However, I haven't been able to reproduce the bug this change is addressing. Here's what I've done, all without the fix from this PR:

- Apply https://github.com/DataDog/integrations-core/commit/4d9ca25adbb17a0d5e92dd572ec364b288a1822e to my local `ddev` installation. This is need so that `DD_APM_ENABLED=true` won't get overridden
- Also set `DD_APM_RECEIVER_SOCKET=""` in `conftest.py`. In the [trace agent configuration](https://github.com/DataDog/datadog-agent/blob/d32089d73949a5c4b09e87a675988112fd6bdee8/pkg/config/config_template.yaml#L1266-L1271) it seems like doing this ought to disable the trace agent UDS listener.
- Ran `ddev env start --dev --agent datadog/agent:7.72.0-rc.1 go_pprof_scraper py3.12` to create a test environment with the agent release candidate where this check seems to not work
- Ran `ddev env test go_pprof_scraper py3.12`

With all these steps, the test passes when I would expect it to fail. Here's what I see in the trace agent logs:

```
2025-10-07 14:02:54 UTC | TRACE | INFO | (pkg/trace/api/api.go:309 in Start) | Listening for traces at http://0.0.0.0:8126
2025-10-07 14:02:54 UTC | TRACE | INFO | (pkg/trace/api/api.go:315 in Start) | Using UDS listener at /var/run/datadog/apm.socket
2025-10-07 14:02:54 UTC | TRACE | INFO | (pkg/trace/api/listener.go:41 in NewMeasuredListener) | Listener started with 1000 maximum connections.
2025-10-07 14:02:54 UTC | TRACE | INFO | (pkg/trace/api/api.go:347 in Start) | Listening for traces at unix:///var/run/datadog/apm.socket
```

and in the environment:

```
% docker exec dd_go_pprof_scraper_py3.12 env | grep DD_APM
DD_APM_ENABLED=true
DD_APM_PROFILING_DD_URL=http://localhost:9999/profiles
DD_APM_RECEIVER_SOCKET=
```

As I understand it, the issue we were hitting with the 7.72.0-rc.1 agent release candidate is that it reports an `apm_config.receiver_socket` default value even if the receiver socket is disabled. [See this commit](https://github.com/DataDog/datadog-agent/commit/5dc07cc3d7a7edca8b25f84c7635a99f00289336). So what I'm trying to do is disable the receiver socket. Yet it seems like the agent using the receiver socket anyway.

So, while the check still seems to work and the changes generally look okay to me, I haven't yet confirmed that this fixes the bug. @Kyle-Neale do you think I may have missed something in my testing?

nsrip-dd (approved):
Okay, if I do something like this:

```
if os.getenv("GO_PPROF_TEST_UDS"):
    E2E_METADATA["env_vars"]["DD_APM_RECEIVER_SOCKET"] = "/var/run/datadog-apm.socket"
else:
    E2E_METADATA["env_vars"]["DD_APM_RECEIVER_SOCKET"] = " an invalid value"
```

and the equivalent with the code prior to this PR, then without this PR the test fails and with it the test succeeds. The failure is a `FileNotFoundError`, like expected. I _think_ the filename actually might be valid, but since it's not an absolute path maybe the check is running from a different working directory and doesn't see the socket? Either way, that gives me more confidence that we'll correctly handle not finding the socket.

It'd be good to have an extra set of eyes on this, since I don't work on this code or the agent much. We could also maybe confirm this in staging?

go_pprof_scraper/datadog_checks/go_pprof_scraper/check.py
```diff
@@ -168,16 +189,70 @@ def add_form_field(name, value):
...
+                    )
+                    last_error = e
+                    # Continue to TCP fallback
+
+            if not upload_successful:
+                try:
+                    r = self.http.post(self.trace_agent_url, files=files)
+                    r.raise_for_status()
+                    # Only set preference if we haven't already determined Unix socket works
+                    if self._preferred_upload_method != "unix":
```
sarah-witt: What cases does this handle?

go_pprof_scraper/tests/test_go_pprof_scraper.py
```diff
@@ -64,6 +65,94 @@ def test_emits_critical_service_check_when_service_is_down(dd_run_check, aggrega
...
+
+        # Create check instance
+        check = GoPprofScraperCheck("go_pprof_scraper", INIT_CONFIG, [instance])
+
+        # Initially, no preference is set
+        assert check._preferred_upload_method is None
+        assert check.trace_agent_socket is not None
+
+        # After first failed Unix socket attempt, should remember to use TCP
+        # (This would be set during check() execution when Unix socket fails)
```
sarah-witt: Should we run the check here?

Kyle-Neale: No, since it'll attempt to reach an endpoint that doesn't exist. We'll also have to do 3 levels of patching to make this work wit check.check() Do you think it's worth having without the `check.check()` being executed? If not, I can remove.

sarah-witt (approved):

PR #2967, opened 2026-04-16
[zabbix] Add Linux: prefix support for Zabbix 6.2+ template item names

shin25818:
## Summary
Zabbix 6.2 (released July 2022) changed the naming convention of template 
item names by adding prefixes such as "Linux: " to item names.

For example:
- Before 6.2: `Load average (1m avg)`
- After 6.2:  `Linux: Load average (1m avg)`

The current METRICS list only contains the old naming convention, causing 
Metric Samples to remain at 0 for all users running Zabbix 6.2 or later.

## Changes
Added `Linux:` prefixed versions of existing metric mappings to support 
Zabbix 6.2+ while maintaining backward compatibility with older versions.

## Testing
Verified with Zabbix 6.4.21 that all item names are now found in metric 
mapping (previously logged as "Item name Linux: XXX not found in metric mapping").

## Impact
- No impact on users running Zabbix 6.0 or earlier
- Fixes metric collection for users running Zabbix 6.2 or later


shin25818:
## Additional Note

While this PR adds support for the `Linux:` prefix, I realized that Zabbix 6.2+ 
also introduced prefixes for other OS templates, such as:

- `Windows: CPU utilization`
- `macOS: CPU utilization`
- etc.

A more robust solution might be to normalize item names by stripping the prefix 
before matching against the METRICS dictionary in `check.py`:

```python
normalized_name = item_name.split(': ', 1)[-1] if ': ' in item_name else item_name
if normalized_name in METRICS:
```

This would handle all OS prefixes without needing to enumerate each one explicitly.
I'm happy to update this PR if the maintainer prefers this approach.

PR #2972, opened 2026-04-21
[zscaler] Add OCSF normalization to all NSS log pipelines

cepolation-datadog:
### What does this PR do?

Adds OCSF 1.5.0 sub-pipelines for every `zscalernss-*` log type so Zscaler logs flow through Datadog's normalized security detection surface.

**Classes mapped (one sub-pipeline each):**

| Log type | OCSF class |
|---|---|
| `zscalernss-web` | HTTP Activity [4002] |
| `zscalernss-fw` | Network Activity [4001] |
| `zscalernss-dns` | DNS Activity [4003] |
| `zscalernss-tunnel` | Tunnel Activity [4014] |
| `zscalernss-emaildlp` (with severity) | Data Security Finding [2006] |
| `zscalernss-endpointdlp` (with severity) | Data Security Finding [2006] |
| `zscalernss-casb` (with severity) | Data Security Finding [2006] (CASB DLP) |
| `zscalernss-casb` (no severity) | File Hosting Activity [6006] |
| `zscalernss-audit` (`LOGIN`/`AUTH` category) | Authentication [3002] |
| `zscalernss-audit` (`USER_MANAGEMENT` or `ROLE_MANAGEMENT`) | Account Change [3001] |
| `zscalernss-audit` (any other category) | API Activity [6003] |
| everything else | Base Event [0] |

**Pre-OCSF helper pipelines** (run on multiple sourcetypes before any class-specific sub-pipeline):

- **`OCSF pre transformations`** — sets `ocsf.metadata.product.{vendor_name,name}`, extracts the sourcetype suffix into `ocsf.metadata.product.feature.name`, normalizes `zscaler.datetime` to epoch-ms in `ocsf.time`, and conditionally builds the `ocsf.file.hashes` array from `zscaler.filemd5` / `zscaler.filesha`.
- **`CASB file name parsing`** — for non-alert CASB logs, maps `zscaler.object_name_1` (or `..._2`) into `ocsf.file.name` when the corresponding `zscaler.object_type_*` is `File`.

**Product metadata:**
- `ocsf.metadata.product.vendor_name` = `"Zscaler"`
- `ocsf.metadata.product.name` = `"ZIA"`
- `ocsf.metadata.product.feature.name` = sourcetype suffix (`web`, `fw`, `dns`, `tunnel`, `casb`, `emaildlp`, `endpointdlp`, `audit`)

Adds the full set of OCSF facets (`ocsf.activity_*`, `ocsf.category_*`, `ocsf.class_*`, `ocsf.status_*`, `ocsf.severity_*`, `ocsf.finding_info.*`, `ocsf.file.*`, `ocsf.tunnel_type`, `ocsf.data_security.*`, etc.) in the canonical `groups/name/path/source` minimal form, reusing names from existing OCSF facets where the path is shared.

### Motivation

Zscaler is one of our most widely used security integrations, but its pipeline didn't emit OCSF yet — meaning Zscaler logs couldn't participate in cross-vendor OCSF-based detection rules, dashboards, or normalization workflows. This change closes that gap.

### Pipeline details

The existing pipeline (remappers, grok parsers, user-agent parser, geoip, status-remapper, etc.) is **unchanged in behavior**. The OCSF sub-pipelines run after all existing processing and consume the already-remapped `zscaler.*` fields.

**Key processor types per OCSF style guide:**
- `pipeline` blocks with `ocsf.isOcsf: true` — one per OCSF class, ordered by ascending `class_uid`, with Base Event [0] as a final catch-all
- `schema-processor` with inline `schema-remapper`, `schema-category-mapper` mappers
- `string-builder-processor` for hardcoded values (product name/vendor, static severities)
- `grok-parser` for type conversions (boolean, integer)
- `attribute-remapper` only used in the existing pre-OCSF section (unchanged) and in the OCSF pre-transformations / CASB-file-name pipelines
- `array-processor` for OCSF list-valued fields (`ocsf.file.hashes`, `ocsf.finding_info.{data_sources,types}`, `ocsf.email.recipients`, etc.)

**Time handling:** `OCSF pre transformations` parses `zscaler.datetime` (always present in Zscaler NSS feeds) directly to `ocsf.time` as epoch-ms via grok. Each sub-pipeline self-maps `ocsf.time` so the `schema-processor` keeps the value.

**Filter exclusivity (no field leakage between classes):**
- DLP sub-pipelines (Email/Endpoint/CASB) include `(-@zscaler.severity:* OR @zscaler.severity:(None OR none))` so non-alert logs route to File Hosting / API Activity instead of Data Security Finding.
- `zscalernss-fw` Network Activity admits `ipsrulelabel`/`threatname` placeholder values `"None"`/`"none"` — the documented Zscaler firewall feed populates these even on non-threat policy events.

**Category-mapper catch-alls (Datadog Other/99 convention):** wildcard catch-alls on string-enum fields use `Other` (id `99`) with a `fallback:` block that copies the source field's literal value into the OCSF `*_name` target — preserving the vendor's label (e.g. `RareCustomAction`) instead of collapsing it to a generic `"Other"`. Numeric-range and absent-field catch-alls keep `Unknown` (id `0`). Applied to: HTTP Activity `activity_id` (from `requestmethod`), File Hosting `activity_id` (from `activity_type_name`), Endpoint DLP `device.type_id` (from `devicetype`), and Endpoint DLP `file.type_id` (from `itemtype`/`filetypecategory`). This clears the OCSF validator's `attribute_enum_sibling_suspicious_other` warning while keeping the catch-all behavior.

**Tunnel device constraint:** `zscaler.sourceip` (the customer-side tunnel endpoint per the Zscaler NSS feed format — `destinationip` is the Zscaler edge VIP via the `%s{destvip}` token) is mapped to `ocsf.device.ip` so Tunnel Activity satisfies its `at_least_one` constraint on `device.{hostname,instance_uid,interface_name,interface_uid,ip,name,uid}`.

**Several existing `attribute-remapper`s** were flipped from `preserveSource: false` to `preserveSource: true` (e.g. `clt_sip` → `network.client.ip`, `srv_dip` → `network.destination.ip`, `dns_req` → `dns.question.name`) so the OCSF sub-pipelines can still read the original `zscaler.*` fields. No existing attribute paths were deleted — `zscaler.*` originals are kept alongside the remapped values.

**Restored byte-facet metadata** — `Total Bytes`, `Zscaler Request Size`, and `Zscaler Response Size` facets keep their `type: integer` + `unit: {family: bytes, name: byte}` so range filtering and byte-formatted display still work in the explorer.

Mappings were generated using Fleak from the existing samples in `zscaler_tests.yaml`, then translated into Datadog pipeline YAML following the [OCSF pipeline style guide](https://datadoghq.atlassian.net/wiki/spaces/~67cef761b4094986b02475ff/pages/5787680769/OCSF+pipeline+style+guide).

### Tests

`zscaler_tests.yaml` has **one test per OCSF sub-pipeline class** (11 tests total). Each test exercises a different class so the pipeline's full breadth is covered without duplication.

| # | sourcetype | OCSF class |
|---|---|---|
| 1 | `zscalernss-web` | HTTP Activity [4002] |
| 2 | `zscalernss-fw` | Network Activity [4001] |
| 3 | `zscalernss-dns` | DNS Activity [4003] |
| 4 | `zscalernss-audit` | API Activity [6003] |
| 5 | `zscalernss-casb` (no severity) | File Hosting Activity [6006] |
| 6 | `zscalernss-tunnel` | Tunnel Activity [4014] |
| 7 | `zscalernss-endpointdlp` | Data Security Finding [2006] |
| 8 | `zscalernss-emaildlp` | Data Security Finding [2006] |
| 9 | `zscalernss-casb` (with severity) | Data Security Finding [2006] (CASB DLP) |
| 10 | `zscalernss-audit` (LOGIN) | Authentication [3002] |
| 11 | `zscalernss-audit` (USER_MANAGEMENT) | Account Change [3001] |

### Review checklist

- [x] PR has a [meaningful title](https://github.com/DataDog/integrations-core/blob/master/CONTRIBUTING.md#pull-request-title)
- [x] Feature has tests (11 test cases — one per OCSF sub-pipeline class — all with refreshed OCSF expected blocks)
- [x] CHANGELOG updated (2.1.0 entry)
- [x] Pipeline impact described above

### Additional Notes

- Validated locally against `logs-backend/domains/event-platform/libs/ocsf-validator-cli`: all 11 tests `VALID`, 100% required OCSF fields, no `attribute_enum_sibling_suspicious_other` warnings.
- Every emitted `ocsf.time` is a real epoch-ms timestamp (no `time: 0` fallbacks).
- `preserveSource: true` flips are backwards compatible — no existing attribute paths were removed.

## Staging Pipeline
https://dd.datad0g.com/logs/pipelines/?search=zscaler

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cepolation-datadog:
@codex review

jbfeldman-dd:
@codex review

zscaler/assets/logs/zscaler.yaml
```diff
@@ -421,18 +493,38 @@ pipeline:
       enabled: true
       sources:
         - sourcetype
+    - type: attribute-remapper
```
jbfeldman-dd: is there a reason you're creating a new field `zscaler.datetime`? We generally try to avoid creating non-OCSF fields

zscaler/assets/logs/zscaler.yaml
```diff
@@ -421,18 +493,38 @@ pipeline:
...
-        - 2021-09-22 17:53:52
-        - Wed Sep 22 17:54:17 2021
+        - "2021-09-22 17:53:52"
+        - "Wed Sep 22 17:54:17 2021"
       grok:
         supportRules: ""
         matchRules: |-
           date_1 %{date("yyyy-MM-dd HH:mm:ss"):date}
           date_2 %{date("EEE MMM dd HH:mm:ss yyyy"):date}
+    - type: grok-parser
```
jbfeldman-dd: Same as above - we usually don't create new fields like `date`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          template: "1"
+          target: ocsf.activity_id
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.activity_name
+          enabled: true
+          template: "Create"
+          target: ocsf.activity_name
+          replaceMissing: false
```
jbfeldman-dd: This is better done is a schema-category-mapper

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          template: "Rule"
+          target: ocsf.finding_info.analytic.type
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.finding_info.analytic.type_id
+          enabled: true
+          template: "1"
+          target: ocsf.finding_info.analytic.type_id
+          replaceMissing: false
```
jbfeldman-dd: Should be done in a schema-category-mapper

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+                  name: High
+                  id: 4
+                - filter:
+                    query: "@zscaler.threatseverity:Critical"
+                  name: Critical
+                  id: 5
+                - filter:
+                    query: "@zscaler.threatseverity:*"
+                  name: Unknown
+                  id: 0
```
jbfeldman-dd: In general, it's better to map to 99/Other rather than unknown

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+        isOcsf: true
+      filter:
+        query: "@sourcetype:zscalernss-alert"
+      processors:
+        - type: string-builder-processor
+          name: Add ocsf.metadata.product.feature.name for alert logs
+          enabled: true
+          template: "alert"
+          target: ocsf.metadata.product.feature.name
+          replaceMissing: false
```
jbfeldman-dd: this shouldn't be needed as it should already be populated by the grok parser in the pre transformation pipeline

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+        - type: grok-parser
+          name: Parse `Activity count` from `description`
+          enabled: true
+          source: description
+          samples:
+            - "Alert ID: 9999999999 | Rule Name: Default Bulk Download of Data Alert | Alert Type: Bulk download alert | User: test@example.com | Activity count: 107 | ZIA-URL: https://admin.zscalertwo.net/"
+          grok:
+            supportRules: ""
+            matchRules: |-
+              activity_count .*?Activity count:\s*%{integer:activity_count}.*
```
jbfeldman-dd: You should map this directly to the OCSF field (`ocsf.finding_info.related_events_count`) rather than using an intermediate field

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          template: "Rule"
+          target: ocsf.finding_info.analytic.type
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.finding_info.analytic.type_id
+          enabled: true
+          template: "1"
+          target: ocsf.finding_info.analytic.type_id
+          replaceMissing: false
```
jbfeldman-dd: Should be done via a schema-category-mapper

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+        - type: grok-parser
+          name: Parse data_sources string into array
+          enabled: true
+          source: ocsf.finding_info.data_sources
+          samples:
+            - "[Zscaler-ZIA-UEBA]"
+          grok:
+            supportRules: ""
+            matchRules: |-
+              data_sources %{data:ocsf.finding_info.data_sources:array("[]",",")}
```
jbfeldman-dd: I would prefer to to use an array-processor for this, as it will only use one processor instead of two

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              targetFormat: integer
+            - type: schema-remapper
+              name: Map `alias` to `ocsf.metadata.log_name`
+              sources:
+                - alias
+              target: ocsf.metadata.log_name
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `alertId` to `ocsf.metadata.event_code`
```
jbfeldman-dd: I don't think `alertId` really corresponds to an event_code. An event_code is usually the name of the rule, not the alert. alertId is more the uid (as it is accurately mapped below)

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          template: "1"
+          target: ocsf.activity_id
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.activity_name
+          enabled: true
+          template: "Create"
+          target: ocsf.activity_name
+          replaceMissing: false
```
jbfeldman-dd: Should be done with a `schema-category-processor`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          template: "Data in-Transit"
+          target: ocsf.data_security.data_lifecycle_state
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.data_security.data_lifecycle_state_id
+          enabled: true
+          template: "2"
+          target: ocsf.data_security.data_lifecycle_state_id
+          replaceMissing: false
```
jbfeldman-dd: Should be done with a schema-category-processor

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          template: "Secure Email Gateway"
+          target: ocsf.data_security.detection_system
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.data_security.detection_system_id
+          enabled: true
+          template: "6"
+          target: ocsf.data_security.detection_system_id
+          replaceMissing: false
```
jbfeldman-dd: Should be done with a schema-category-processor

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.user` to `ocsf.actor.user.name`
+              sources:
+                - zscaler.user
+              target: ocsf.actor.user.name
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.itemname` to `ocsf.file.name`
```
jbfeldman-dd, line 1715: In the sample logs, `itemname` isn't a file

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              name: ocsf.data_security.data_lifecycle_state_id
+              categories:
+                - filter:
+                    query: "@zscaler.activitytype:CLIPBOARD_PASTE"
+                  name: Data in-Use
+                  id: 3
+                - filter:
+                    query: "@zscaler.channel:(CLOUD_UPLOAD OR NETWORK_SHARE OR REMOVABLE_STORAGE)"
+                  name: Data in-Transit
+                  id: 2
```
jbfeldman-dd, line 1835: is this is an exhaustive list of `activitytype`s? It also doesn't match the test log activity_type, which is `email_sent`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+                - zscaler.threatname
+                - zscaler.rulelabel
+              target: ocsf.message
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-category-mapper
+              name: ocsf.severity_id
+              categories:
+                - filter:
+                    query: "@zscaler.threatseverity:Informational"
```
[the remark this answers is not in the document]
cepolation-datadog: done

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              sources:
+                - zscaler.cdfqdn
+              target: ocsf.dst_endpoint.hostname
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.ipproto` to `ocsf.connection_info.protocol_name`
+              sources:
+                - zscaler.ipproto
+              target: ocsf.connection_info.protocol_name
```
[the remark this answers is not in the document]
cepolation-datadog: done

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+                - ocsf.data_security.data_lifecycle_state_id
+              target: ocsf.data_security.data_lifecycle_state_id
+              preserveSource: true
+              overrideOnConflict: true
+              targetFormat: integer
+            - type: schema-category-mapper
+              name: ocsf.severity_id
+              categories:
+                - filter:
+                    query: "@zscaler.severity:Informational"
```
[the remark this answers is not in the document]
cepolation-datadog: done

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+                  id: 1
+                - filter:
+                    query: "@zscaler.result:FAILURE"
+                  name: Medium
+                  id: 3
+                - filter:
+                    query: "@zscaler.result:*"
+                  name: Unknown
+                  id: 0
+              targets:
```
jbfeldman-dd, line 2396: Generally we just pass through the severity of the original log source rather than assigning our own based on criteria like success/failure.

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          template: "true"
+          target: ocsf.is_remote
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.logon_type
+          enabled: true
+          template: "%{zscaler.interface}"
+          target: ocsf.logon_type
+          replaceMissing: false
```
jbfeldman-dd: This should be a `schema-remapper` in the `schema-processor`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.adminid` to `ocsf.actor.user.email_addr`
+              sources:
+                - zscaler.adminid
+              target: ocsf.actor.user.email_addr
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.resource` to `ocsf.user.name`
```
jbfeldman-dd: in the sample log, `resource` refers to the role being assigned when it's a `ROLE_MANAGEMENT` event

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+            - type: schema-remapper
+              name: Map `ocsf.logon_type_id` to `ocsf.logon_type_id`
+              sources:
+                - ocsf.logon_type_id
+              target: ocsf.logon_type_id
+              preserveSource: true
+              overrideOnConflict: true
+              targetFormat: integer
+            - type: schema-remapper
+              name: Map `zscaler.adminid` to `ocsf.user.name`
```
jbfeldman-dd, line 2465: The subcategory for the sample log we have is ADMIN_LOGIN, which is why the `adminid` field is the user. For non-admin login events, the field for logins might be different, and it might be worth investigating in NBT

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.clientip` to `ocsf.src_endpoint.ip`
+              sources:
+                - zscaler.clientip
+              target: ocsf.src_endpoint.ip
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.recordid` to `ocsf.session.uid`
```
jbfeldman-dd: This is better mapped to `ocsf.metadata.uid`, as I don't believe the `recordid` is a value that persists across a login session

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.dstip_country` to `ocsf.dst_endpoint.location.country`
+              sources:
+                - zscaler.dstip_country
+              target: ocsf.dst_endpoint.location.country
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.hostname` to `ocsf.http_request.url.hostname`
```
jbfeldman-dd, line 3003: hostname should also be mapped to `ocsf.dst_endpoint.hostname`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.status` to `ocsf.status_code`
+              sources:
+                - zscaler.status
+              target: ocsf.status_code
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.event_id` to `ocsf.http_request.uid`
```
jbfeldman-dd: This should be mapped to `ocsf.metadata.uid` instead

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.url` to `ocsf.http_request.url.url_string`
+              sources:
+                - zscaler.url
+              target: ocsf.http_request.url.url_string
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.protocol` to `ocsf.http_request.url.scheme`
```
jbfeldman-dd: In the sample log, the protocol is "SSL", which doesn't make sense as a schema. I'd remove this processor

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          target: ocsf.device.type
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.device.type_id
+          enabled: true
+          template: "0"
+          target: ocsf.device.type_id
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Map `zscaler.tunneltype` to `ocsf.tunnel_type`
```
jbfeldman-dd: Is this a typo/mistake? Name says is maps from a source field, but it hardcodes as "unknown"

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+            - type: schema-remapper
+              name: Map `zscaler.dstport` to `ocsf.dst_endpoint.port`
+              sources:
+                - zscaler.dstport
+              target: ocsf.dst_endpoint.port
+              preserveSource: true
+              overrideOnConflict: true
+              targetFormat: integer
+            - type: schema-remapper
+              name: Map `zscaler.tunneltype` to `ocsf.tunnel_type`
```
jbfeldman-dd: If the tunnel type is going to be mapped, then type_id should not be hardcodes to `None`. Unless you have a good mapping of Zscaler tunnel types to the OCSF standard of "Split/Full", it should be removed

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+                id: ocsf.activity_id
+            - type: schema-category-mapper
+              name: ocsf.status_id
+              categories:
+                - filter:
+                    query: "@zscaler.event:TUNNEL_UP OR @zscaler.tunnelactionname:TUNNEL_UP"
+                  name: Success
+                  id: 1
+                - filter:
+                    query: "@zscaler.event:TUNNEL_DOWN OR @zscaler.tunnelactionname:TUNNEL_DOWN"
```
jbfeldman-dd: is a tunnel closing a failure? Not sure this is a correct mapping

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+            - type: schema-category-mapper
+              name: ocsf.severity_id
+              categories:
+                - filter:
+                    query: "@zscaler.eventreason:PeerUnreachable"
+                  name: Medium
+                  id: 3
+                - filter:
+                    query: "@zscaler.event:TUNNEL_DOWN OR @zscaler.tunnelactionname:TUNNEL_DOWN"
+                  name: Low
```
jbfeldman-dd: Generally we just pass through severity or set it to "Informational" rather than trying to synthesize it through event outcomes

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+            classUid: 4014
+            extensions: []
+            profiles: []
+    - type: pipeline
+      name: OCSF sub pipeline for class API Activity [6003]
+      enabled: true
+      ocsf:
+        isOcsf: true
+      filter:
+        query: "@sourcetype:zscalernss-audit AND -@zscaler.category:(AUTH OR LOGIN OR USER_MANAGEMENT OR ROLE_MANAGEMENT)"
```
jbfeldman-dd, line 3574: The category AUTH is excluded here, but not included in the Authentication sub pipeline or any other sub-pipelines. It should be mapped somewhere unless there's a reason not to

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+      processors:
+        - type: string-builder-processor
```
jbfeldman-dd: This should be a `schema-remapper`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          name: Add ocsf.api.operation
+          enabled: true
+          template: "%{zscaler.action}"
+          target: ocsf.api.operation
+          replaceMissing: false
+        - type: attribute-remapper
+          name: Map `zscaler.resource` to `ocsf.resource.uid`
```
jbfeldman-dd: This is a name, not a uid, this mapper should be removed

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+          enabled: true
+          sources:
+            - zscaler.resource
+          sourceType: attribute
+          target: ocsf.resource.name
+          targetType: attribute
+          preserveSource: true
+          overrideOnConflict: false
+        - type: attribute-remapper
+          name: Map `zscaler.category` to `ocsf.resource.type`
```
jbfeldman-dd: `zscaler.subcategory` is a better mapping than `zscaler.category`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              overrideOnConflict: true
+              targetFormat: integer
+            - type: schema-remapper
+              name: Map `zscaler.tenant` to `ocsf.metadata.tenant_uid`
+              sources:
+                - zscaler.tenant
+              target: ocsf.metadata.tenant_uid
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
```
jbfeldman-dd, line 3811: In the example log we have, the application name is the SaaS application the user is interacting with, not the user itself. Might be more appropriate to map to `ocsf.dst_endpoint.name`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              name: Map `zscaler.hostname`, `zscaler.tenant` to `ocsf.src_endpoint.hostname`
+              sources:
+                - zscaler.hostname
+                - zscaler.tenant
+              target: ocsf.src_endpoint.hostname
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.extownername`, `zscaler.external_owner_name` to `ocsf.dst_endpoint.name`
```
jbfeldman-dd: The logs we have show `extownername` as the external owner of a file, not the name of an application or network endpoint. This is better mapped to `ocsf.file.owner.name`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+                - filter:
+                    query: "@zscaler.activity_type_name:FILE_SHARED_EXTERNAL"
+                  name: Share
+                  id: 12
+                - filter:
+                    query: "@zscaler.activity_type_name:LOGIN_SUCCESS"
+                  name: Access
+                  id: 3
+                - filter:
+                    query: "@sourcetype:zscalernss-casb"
```
jbfeldman-dd, line 3917: These should be in numeric order, unless otherwise needed for control flow

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1222,3 +1314,3119 @@ pipeline:
...
+              name: ocsf.status_id
+              categories:
+                - filter:
+                    query: "@sourcetype:zscalernss-casb"
+                  name: Success
+                  id: 1
+              targets:
+                name: ocsf.status
+                id: ocsf.status_id
+            - type: schema-category-mapper
```
jbfeldman-dd, line 3947: Default severity levels are already set by the pre-transformation pipeline

jbfeldman-dd (commented):
Left comments. Here are a common patterns that should be fixed
    - In general, we prefer to map to 99/Other unless the field is truly unknown
    - In several places, you would hardcode a an enum name/pair (like activity_id) using string-builder-processors. It's better to use a `schema-category-mapper` for all enum pairs, even if there's only one possible value. It's more compact and better leverages the schema-processor
    - When possible, map to `ocsf.medata.event_code`. This is OCSF's equivalent of `evt.name`, and is a major way we filter logs
    - The creation of the fields `zscaler.datetime` and `date` is should be removed. Instead, use a grok-parser to directly map to `ocsf.time` or other time fields in the pre transformations pipeline. Creation of intermediate fields should be avoided unless there is a specific reason to do so.
    - Avoid using string-builder-processors if just copying a field. Use a schema-remapper instead.

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: The pipeline below was looking for this field and it wasn't in the pipeline. Claude tried to fix it.

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: field was needed to be converted to ms, how do we normally do these kinds of fixes?

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: @jbfeldman-dd  would it be better to not map this at all since it can be different things? I couldn't find anything better in customer data when I looked.

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: No, documentation doesn't appear to have an exhaustive list, how should we treat it?

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: You sure? The ZScaler docs have it as this, it's a web request too. Wouldn't a destination hostname be different? Or do you mean map it to that in addition to what I have it mapped to?
image
https://help.zscaler.com/zia/nss-feed-output-format-web-logs

cepolation-datadog:
@codex review

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+              preserveSource: true
+              overrideOnConflict: false
+            - type: array-processor
+              name: Append `tmp_sha256` to `ocsf.file.hashes`
+              enabled: true
+              operation:
+                source: tmp_sha256
+                target: ocsf.file.hashes
+                preserveSource: false
+                type: append
```
jbfeldman-dd: This logic can be done without having to use sub pipelines. Just use `category-processor` in the relevant pipeline(s). Don't create a fallback where `zscaler.filesha` or `zscaler.filemd5` don't exist, which will leave the tmp fields unpopulated and result in no-ops

The current implementation also creates validation errors in hashes exist in data sources that are mapped to classes that have an `ocsf.file` attribute (like Base Event)

Example implementation

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+    - type: pipeline
+      name: OCSF sub pipeline for class Detection Finding [2004] (Alert)
+      enabled: true
```
jbfeldman-dd, line 1431: I meant to mention this in my last review, but I would strongly consider removing this sub pipeline and folding these events into the 4001 and 4002 pipelines with the `security_control` profile. This enables these events to better map network fields like source and destination endpoints. Attributes added by the `security_control` profile are all optional, so can be ignored if there is no `zscaler.threatseverity` associated with the event

If you don't want to migrate it to 4001 and 4002, you should map observables like src and destination ip under `ocsf.evidences` array

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+            extensions: []
+            profiles: []
+    - type: pipeline
+      name: OCSF sub pipeline for class Detection Finding [2004] (Firewall Threat)
+      enabled: true
+      ocsf:
+        isOcsf: true
+      filter:
+        query: "@sourcetype:zscalernss-fw AND ((@zscaler.ipsrulelabel:* -@zscaler.ipsrulelabel:(None OR none)) OR (@zscaler.threatname:* -@zscaler.threatname:(None OR none))) AND (-@zscaler.threatseverity:* OR @zscaler.threatseverity:(None OR none))"
+      processors:
```
jbfeldman-dd, line 1373: Same comment as above - consider rolling into existing 4001 pipeline

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: Yeah, removing the mapping is the right call probably

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: It looks like you ended up finding a good list. I'd remove the comment, otherwise looks solid

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+          template: "Regular File"
+          target: ocsf.file.type
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.file.type_id default
+          enabled: true
+          template: "1"
+          target: ocsf.file.type_id
+          replaceMissing: false
+        - type: schema-processor
```
jbfeldman-dd, line 2014: This enum should be a `schema-category-mapper`, not `string-builder-processor`s

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `ocsf.file.type_id` to `ocsf.file.type_id`
+              sources:
+                - ocsf.file.type_id
+              target: ocsf.file.type_id
+              preserveSource: true
+              overrideOnConflict: true
+              targetFormat: integer
```
jbfeldman-dd: should be removed in favor of the above mentioned `schema-category-processor`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+          enabled: true
+          template: "Unknown"
+          target: ocsf.connection_info.direction
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.connection_info.direction_id
+          enabled: true
+          template: "0"
+          target: ocsf.connection_info.direction_id
+          replaceMissing: false
```
jbfeldman-dd: This enum pair should be migrated to `schema-category-mapper`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+          enabled: true
+          template: "6"
+          target: ocsf.activity_id
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.activity_name
+          enabled: true
+          template: "Traffic"
+          target: ocsf.activity_name
+          replaceMissing: false
```
jbfeldman-dd: This enum pair should be migrated to schema-category-mappe

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+          name: Add ocsf.connection_info.direction
+          enabled: true
+          template: "Unknown"
+          target: ocsf.connection_info.direction
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.connection_info.direction_id
+          enabled: true
+          template: "0"
+          target: ocsf.connection_info.direction_id
```
jbfeldman-dd: Another enum pair to migrate to a `schema-category-mapper`

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: You should map both I think. Your current mapping is correct, it should be `ocsf.dst_endpoint.hostname` as well

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+          template: "Response"
+          target: ocsf.activity_name
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.activity_id
+          enabled: true
+          template: "2"
+          target: ocsf.activity_id
+          replaceMissing: false
+        - type: attribute-remapper
```
jbfeldman-dd, line 3237: Another enum pair to migrate to a`schema-category-mapper`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+              overrideOnConflict: true
+              targetFormat: integer
+            - type: schema-remapper
+              name: Map `ocsf.severity_id` to `ocsf.severity_id`
+              sources:
+                - ocsf.severity_id
+              target: ocsf.severity_id
+              preserveSource: true
+              overrideOnConflict: true
+              targetFormat: integer
```
jbfeldman-dd: This isn't mapped upstreamed, so no need to copy it to itself, so it should be removed. I think you maybe meant to migrate the previous mapping from a s`tring-builder-processor` to a s`chema-category-mapper`, but deleted the hardcoded mappings without adding the new ones

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+          enabled: true
+          template: "Unknown"
+          target: ocsf.device.type
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.device.type_id
+          enabled: true
+          template: "0"
+          target: ocsf.device.type_id
+          replaceMissing: false
```
jbfeldman-dd: Migrate to `schema-category-mapper`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+              target: ocsf.device.name
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `zscaler.sourceip` to `ocsf.device.ip`
+              sources:
+                - zscaler.sourceip
+              target: ocsf.device.ip
+              preserveSource: true
+              overrideOnConflict: true
```
jbfeldman-dd: I'm not sure this device mapping is really accurate - I don't think QUICKSEC is really a reporting device, and the IPs are already mapped to `src_endpoint.ip`. Can probably just remove these mappings without issue

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+                    query: "@zscaler.event:TUNNEL_UP OR @zscaler.tunnelactionname:TUNNEL_UP"
+                  name: Open
+                  id: 1
+                - filter:
+                    query: "@zscaler.event:TUNNEL_DOWN OR @zscaler.tunnelactionname:TUNNEL_DOWN"
+                  name: Close
+                  id: 2
+                - filter:
+                    query: "@sourcetype:zscalernss-tunnel"
+                  name: Unknown
```
jbfeldman-dd: This should be a 99 with a fallback do `@zscaler.event`. Also, these fields don't exist in the sample - do they actually exist or are they hallucinated by fleak?

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+              name: ocsf.status_id
+              categories:
+                - filter:
+                    query: "@zscaler.event:TUNNEL_UP OR @zscaler.tunnelactionname:TUNNEL_UP"
+                  name: Success
+                  id: 1
+                - filter:
+                    query: "@sourcetype:zscalernss-tunnel"
+                  name: Unknown
+                  id: 0
```
jbfeldman-dd: Same comment as above

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+            - type: schema-remapper
+              name: Map `zscaler.errorcode` to `ocsf.status_detail`
+              sources:
+                - zscaler.errorcode
+              target: ocsf.status_detail
+              preserveSource: true
+              overrideOnConflict: true
+            - type: schema-remapper
+              name: Map `ocsf.severity_id` to `ocsf.severity_id`
+              sources:
```
jbfeldman-dd: Similar to above, no upstream severity mapping so this doesn't make sense. Resolve by mapping it to a `schema-category-mapper`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,3166 @@ pipeline:
...
+          enabled: true
+          template: "unknown"
+          target: ocsf.file.name
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.file.type
+          enabled: true
+          template: "Regular File"
+          target: ocsf.file.type
+          replaceMissing: false
```
jbfeldman-dd: Migrate to a `schema-category-mapper`

jbfeldman-dd (commented):
Some more comments. Some overall themes
- There's still some more enums being set by string-builder-processors that need to be migrated to schema-category-mappers
- As we discussed over slack, schema-category-mappers that have a fallback case of 99/Other need to have a fallback field to read from (usually the field you're mapping from)

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: I went back and forth and now back again on the 2004 -> 4001 and decided to not make them 2004s. Looking at the data, there is no easy way to identify something as an alert as even the allow things have *rule* key names for things “allowed” in the data I was looking at. So just going to axe them.

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: hallucinated, good catch

zscaler/assets/logs/zscaler.yaml
```diff
@@ -404,40 +483,28 @@ pipeline:
...
-          date_2 %{date("EEE MMM dd HH:mm:ss yyyy"):date}
+          date_1 %{date("yyyy-MM-dd HH:mm:ss"):ocsf.time}
+          date_2 %{date("EEE MMM dd HH:mm:ss yyyy"):ocsf.time}
     - type: date-remapper
-      name: Define `date` as the official date of the log
+      name: Define `ocsf.time` as the official date of the log
       enabled: true
       sources:
-        - date
+        - ocsf.time
```
jbfeldman-dd: This should be moved to the OCSF pre-transformation pipeline

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: reupping this comment, we discussed it over slack as well

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: Bumping this comment

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+              name: ocsf.activity_id
+              categories:
+                - filter:
+                    query: "@sourcetype:zscalernss-casb"
+                  name: Create
+                  id: 1
+                - filter:
+                    query: "*"
+                  name: Other
+                  id: 99
```
jbfeldman-dd: This is nitpicky and doesn't really need to be changed, but in cases where the first filter should match all events in the sub pipeline, you don't need a 99/Other fallback. But not really important

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: Bumping this - did you ever look at it? I'd consider it non-blocking

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                - filter:
+                    query: "@zscaler.action:(Block OR Drop OR \"IPS Block\" OR \"IPS Drop\")"
+                  name: Failure
+                  id: 2
+                - filter:
+                    query: "@sourcetype:zscalernss-fw"
+                  name: Other
+                  id: 99
+              targets:
+                name: ocsf.status
```
jbfeldman-dd, line 2814: Add a fallback for Other/99 here

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                    query: "@zscaler.deviceostype:macOS*"
+                  name: macOS
+                  id: 300
+                - filter:
+                    query: "@zscaler.deviceostype:iOS*"
+                  name: iOS
+                  id: 301
+                - filter:
+                    query: "@sourcetype:zscalernss-fw"
+                  name: Other
```
jbfeldman-dd, line 2848: Add a fallback for Other/99

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                  name: Medium
+                  id: 3
+                - filter:
+                    query: "@zscaler.threatseverity:low"
+                  name: Low
+                  id: 2
+                - filter:
+                    query: "*"
+                  name: Informational
+                  id: 1
```
jbfeldman-dd, line 2907: this is inconsistent with how you've addressed severity in other sub-pipelines. Either they should all have a fallback to the zscaler severity level, or all fallback to Informational

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                - filter:
+                    query: "@zscaler.action:Blocked"
+                  name: Failure
+                  id: 2
+                - filter:
+                    query: "@zscaler.action:*"
+                  name: Other
+                  id: 99
+              targets:
```
jbfeldman-dd, line 3193: Add fallback for 99

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: Un-resolving this: still not migrated

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                - filter:
+                    query: "@zscaler.reqaction:Allow AND @zscaler.resaction:Allow"
+                  name: Success
+                  id: 1
+                - filter:
+                    query: "@zscaler.reqaction:Block OR @zscaler.resaction:Block"
+                  name: Failure
+                  id: 2
+                - filter:
+                    query: "@zscaler.reqaction:* OR @zscaler.resaction:*"
```
jbfeldman-dd, line 3359: Add fallback for 99

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+        - type: attribute-remapper
+          name: Map `zscaler.action` to `ocsf.api.operation`
+          enabled: true
+          sources:
+            - zscaler.action
+          sourceType: attribute
+          target: ocsf.api.operation
+          targetType: attribute
+          preserveSource: true
+          overrideOnConflict: false
```
jbfeldman-dd: This should be a `schema-remapper`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                  id: 3
+                - filter:
+                    query: "@zscaler.action:DELETE"
+                  name: Delete
+                  id: 4
+                - filter:
+                    query: "@sourcetype:zscalernss-audit"
+                  name: Other
+                  id: 99
+              targets:
```
jbfeldman-dd, line 3725: Add fallback case for 99

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                - filter:
+                    query: "@zscaler.result:FAILURE"
+                  name: Failure
+                  id: 2
+                - filter:
+                    query: "@sourcetype:zscalernss-audit"
+                  name: Other
+                  id: 99
+              targets:
+                name: ocsf.status
```
jbfeldman-dd, line 3751: Add fallback case for 99

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+          enabled: true
+          template: "0"
+          target: ocsf.activity_id
+          replaceMissing: false
+        - type: string-builder-processor
+          name: Add ocsf.activity_name
+          enabled: true
+          template: "Unknown"
+          target: ocsf.activity_name
+          replaceMissing: false
```
jbfeldman-dd: This enum should be migrated to a `schema-category-mapepr`

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+                  name: Success
+                  id: 1
+                - filter:
+                    query: "@zscaler.result:FAILURE"
+                  name: Failure
+                  id: 2
+                - filter:
+                    query: "@zscaler.result:*"
+                  name: Other
+                  id: 99
```
jbfeldman-dd, line 2378: Add fallback value for 99

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2818 @@ pipeline:
...
+            - type: schema-category-mapper
+              name: ocsf.logon_type_id
+              categories:
+                - filter:
+                    query: "*"
+                  name: Other
+                  id: 99
+              targets:
+                name: ocsf.logon_type
+                id: ocsf.logon_type_id
```
jbfeldman-dd: If this doesn't have a mapping, I'd just remove it as it's not required

jbfeldman-dd (commented):
Still a few unresolved comments from the last round of review + some cases where there's no fallback for 99/Other

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: Ah claude got me, was suppose to just be assigning datetime to the official timestamp and the grok parser in the pre transformation

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: Image it's needed. Since it deals with data security it would bad to not have file info so I think we leave it as we discussed in slack.

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: yes, from what I can see in documentation, only admins login

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: yeah, had it in a commit I forgot to push, my bad!

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: yourway won’t work, it needs the subpipelines at a minimum if we want to map md5 and sha256, your way(without subpipelines) will only map whichever processor is first

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2804 @@ pipeline:
...
+                    query: "@zscaler.action:SIGN_OUT"
+                  name: Logoff
+                  id: 2
+                - filter:
+                    query: "@zscaler.action:*"
+                  name: Other
+                  id: 99
+              targets:
+                name: ocsf.activity_name
+                id: ocsf.activity_id
```
jbfeldman-dd, line 2552: Needs an other/99 fallback

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1145,80 +1212,2804 @@ pipeline:
...
+                    query: "@zscaler.result:FAILURE"
+                  name: Failure
+                  id: 2
+                - filter:
+                    query: "@zscaler.result:*"
+                  name: Other
+                  id: 99
+              targets:
+                name: ocsf.status
+                id: ocsf.status_id
```
jbfeldman-dd, line 2577: Needs an other/99 fallback

zscaler/assets/logs/zscaler.yaml
jbfeldman-dd: reupping this

jbfeldman-dd (dismissed):
Just a few missing fallbacks, but otherwise looks solid

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1231,6 +1231,138 @@ pipeline:
...
+              name: Map `zscaler.object_name_1` to `ocsf.file.name`
+              enabled: true
+              sources:
+                - zscaler.object_name_1
+              sourceType: attribute
+              target: ocsf.file.name
+              targetType: attribute
+              preserveSource: true
+              overrideOnConflict: false
+        - type: pipeline
```
jbfeldman-dd: I thought this sub pipeline nesting wasn't allowed?

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1231,6 +1231,138 @@ pipeline:
...
+              enabled: true
+              sources:
+                - zscaler.object_name_2
+              sourceType: attribute
+              target: ocsf.file.name
+              targetType: attribute
+              preserveSource: true
+              overrideOnConflict: false
+    - type: pipeline
+      name: Build ocsf.file.hashes MD5 entry
```
jbfeldman-dd, line 1271: need to add isOcsf:true

zscaler/assets/logs/zscaler.yaml
```diff
@@ -1231,6 +1231,138 @@ pipeline:
...
+        - type: array-processor
+          name: Append `tmp_md5` to `ocsf.file.hashes`
+          enabled: true
+          operation:
+            source: tmp_md5
+            target: ocsf.file.hashes
+            preserveSource: false
+            type: append
+    - type: pipeline
+      name: Build ocsf.file.hashes SHA-256 entry
```
jbfeldman-dd, line 1319: add isOcsf:true

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: fixed

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: done

zscaler/assets/logs/zscaler.yaml
cepolation-datadog: done

jbfeldman-dd (approved):

BoyangHuang (approved):

PR #3024, opened 2026-06-02
cloudsmith: add `billable_only` filter to analytics metrics (default: True)

jtaylor-cs:
### What does this PR do?

Adds a `billable_only` boolean configuration option (default: `True`) to the Cloudsmith integration. When `True`, the analytics API is queried with `billable_only=True`, limiting metrics to billable download requests only (download domain, no error responses >= 400). When `False`, all traffic is returned with no filtering applied — equivalent to omitting the parameter entirely.

**Files changed:**

| File | Change |
|---|---|
| `check.py` | Read `billable_only` from instance config (default `True`); append `billable_only=<value>` to every `_build_analytics_url` call |
| `spec.yaml` | Declare `billable_only` as an optional boolean parameter |
| `conf.yaml.example` | Document `billable_only` with `@param` annotation |
| `__about__.py` | Bump version `1.3.0` → `1.4.0` |
| `CHANGELOG.md` | Add v1.4.0 release notes |
| `tests/test_cloudsmith.py` | Add 2 new tests: default→`billable_only=True` in URL; explicit `False`→`billable_only=False` in URL |

### Motivation

Cloudsmith is introducing **native logs** — a new category of traffic that is non-billable and was not previously tracked. Once native logs are live, the analytics API will return a mix of billable and non-billable requests unless filtered. Without this change, customers and Cloudsmith users monitoring download/bandwidth metrics via this Datadog integration would see a sudden and unexplained increase in metric values.

By defaulting `billable_only=True`, the integration maintains the same billing-consistent behaviour it had before native logs were introduced — no customer action required. Users who explicitly want to include all traffic (billable and non-billable) can opt in by setting `billable_only: false`.

### Cloudsmith API parameter details

- **Parameter name:** `billable_only`
- **`True`:** Filters to billable requests only — requests through the download domain with HTTP status < 400
- **`False`:** Returns all results with no filtering applied (includes native CDN logs and errored requests); equivalent to omitting the parameter entirely

### Review checklist

- [x] PR has a meaningful title
- [x] Feature has tests (2 new tests covering default and explicit-false cases; all 48 tests pass)
- [x] Git history is clean (single focused commit)

### Additional Notes

- The `billable_only` query parameter was added to the Cloudsmith Analytics v2 API as part of internal ticket ENG-11511.
- The default of `True` is intentional and backwards-compatible — it preserves existing metric values for all current users.
- This change affects only the analytics time-series endpoint (`/analytics/metrics/client/time-series/`); all other API calls (quota, audit log, vulnerabilities, members, repositories) are unaffected.


cloudsmith/assets/configuration/spec.yaml
```diff
@@ -71,4 +71,10 @@ files:
               - production
             package_format:
               - python
+    - name: billable_only
+      required: false
+      description: Limit analytics to billable downloads only, excluding error responses. Defaults to true.
```
BartoszBlizniak: Small accuracy thing: the billable filter on our side does two things — it restricts to the download domain (request_type='dl') and drops http_status >= 400. Could we reword to something like:

"Limit analytics to billable download requests only (download-domain requests with HTTP status < 400). When false, all traffic — including native CDN logs and errored requests — is returned."

Same applies to the `## @param` description in `conf.yaml.example` (line 62) — worth keeping them in sync.

BartoszBlizniak (commented):
One small comment from me

cloudsmith/assets/configuration/spec.yaml
jtaylor-cs: Updated both spec.yaml and conf.yaml.example with this suggestion.

jtaylor-cs: There is a character limit for the description (90 chars). Had to adjust to successfully pass validations.

jtaylor-cs: I think I figured out a way to work with the longer text description.

joepeeples (approved):

jtaylor-cs:
@london-wharton this PR is still pending your review. Thank you.

BartoszBlizniak:
Hey @dkirov-dd (apologies for direct @ ), but is there any chance we could please push this forward up in your review queue - it's a small parameter change, however, it's a very necessary change on our end as our data team is changing how the API works and this is currently blocking us. Thank you 🙏

jtaylor-cs:
@london-wharton this is still pending your review. Unfortunately, one of the checks is now failing:
Asset Review: Failed
A system error occurred while assets were being validated. 
Contact Datadog to find out how to resolve the problem.

jtaylor-cs:
@joepeeples, can you help get a second reviewer on this pending PR?

steveny91 (approved):

PR #2939, opened 2026-03-20
redis_sentinel: Add SSL/TLS, username/ACL, and Redis 7+ replica support

amendez-primer:
## Summary

Modernizes the Redis Sentinel check to support TLS-enabled deployments, Redis 6+ ACLs, and Redis 7+ command changes.

Fixes #2938, #2395, #1586

### 1. SSL/TLS support (#2938)
The check previously created `redis.StrictRedis` connections without any SSL parameters, making it impossible to connect to TLS-enabled Sentinel instances (e.g. AWS ElastiCache, Azure Cache for Redis).

Added support for: `ssl`, `ssl_certfile`, `ssl_keyfile`, `ssl_ca_certs`, `ssl_cert_reqs` — matching the core `redisdb` integration.

### 2. Username / ACL support (#2395)
Redis 6+ introduced ACLs requiring a `username` parameter. Without it, `AUTH` fails with: `AUTH <password> called without any password configured for the default user`.

Added `sentinel_username` instance config option, passed through to `StrictRedis(username=...)`.

### 3. Redis 7+ REPLICAS command (#1586)
Redis 7 removed the `SENTINEL SLAVES` subcommand in favor of `SENTINEL REPLICAS`. The check now uses `sentinel_replicas()` (redis-py 4.2+) with a fallback to `sentinel_slaves()` for older Redis versions.

### 4. Socket timeout
Added `socket_timeout` config option (default: 5s) to prevent the check from hanging indefinitely on unreachable sentinels.

## Files changed

| File | Change |
|------|--------|
| `redis_sentinel.py` | SSL, username, timeout params in `_load_config()`; `_get_sentinel_replicas()` helper |
| `conf.yaml.example` | Documented all new config options |
| `test_redis_sentinel.py` | Tests for SSL, username, timeout config; updated mock for replicas |
| `pyproject.toml` | Bumped `redis` dep from `2.10.5` to `>=4.5.0` |

## Example config (Helm)
```yaml
datadog:
  confd:
    redis_sentinel.yaml: |-
      instances:
        - sentinel_host: sentinel.example.com
          sentinel_port: 26379
          sentinel_username: default
          sentinel_password: secret
          ssl: true
          ssl_ca_certs: /etc/ssl/certs/ca.pem
          ssl_certfile: /etc/ssl/certs/client.pem
          ssl_keyfile: /etc/ssl/certs/client-key.pem
          socket_timeout: 10
          masters:
            - mymaster
```

## Backward compatibility
[...]

Issue #2395, opened 2024-05-25
redis_sentinel


melissa-hale:
**Output of the [info page](https://docs.datadoghq.com/agent/guide/agent-commands/#service-status)**

When running the redis_sentinel integration to monitor a group of Redis Sentinel servers, the check fails with the following error:

```text
2024-05-24 22:57:33 UTC | CORE | WARN | (pkg/collector/python/datadog_agent.go:131 in LogMessage) | redis_sentinel:e8118be2e61e206d | (redis_sentinel.py:42) | Error collecting metrics for master myprimary: AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?
```

This is my conf.yaml:

```yaml
init_config:

instances:
  - sentinel_host: ${{SENTINEL1_PRIVATE_DOMAIN}}
    sentinel_port: 26379
    sentinel_password: ${{REDIS_PRIMARY_PASSWORD}}
    masters:
      - ${{REDIS_PRIMARY_NAME}}
    tags:
      - instance:redis_sentinel_1
  - sentinel_host: ${{SENTINEL2_PRIVATE_DOMAIN}}
    sentinel_port: 26379
    sentinel_password: ${{REDIS_PRIMARY_PASSWORD}}
    masters:
      - ${{REDIS_PRIMARY_NAME}}
    tags:
      - instance:redis_sentinel_2
  - sentinel_host: ${{SENTINEL3_PRIVATE_DOMAIN}}
    sentinel_port: 26379
    sentinel_password: ${{REDIS_PRIMARY_PASSWORD}}
    masters:
      - ${{REDIS_PRIMARY_NAME}}
    tags:
      - instance:redis_sentinel_3
```

**Additional environment details (Operating System, Cloud provider, etc):**
- Deployed in a Docker environment
- Redis images tried: `redis:latest` & `redis:6.2.6`
- Agent image: `datadog/agent:7`
- Reproduced issue locally on Mac Sonoma

**Steps to reproduce the issue:**
1. Deploy a Redis replica set with Sentinel (followed [this guide](https://redis.io/learn/operate/redis-at-scale/high-availability/introduction))
2. Deploy a DataDog agent
3. Install integtation: `datadog-agent integration install -t datadog-redis_sentinel==1.1.0` (or 1.1.1) in agent container
4. Configure the redis_sentinel check similar to above and restart agent
5. Observe error upon check execution

**Describe the results you received:**
Received an AUTH error during check execution

[...]

Issue #1586, opened 2022-10-31
Redis sentinel integration needs an update because of removed SLAVES command


Akkowicz:
**Output of the [info page](https://docs.datadoghq.com/agent/guide/agent-commands/#service-status)**

```text
    redis_sentinel (1.1.0)
    ----------------------
      Instance ID: redis_sentinel:90bad44364c7949d [WARNING]
      Configuration Source: file:/etc/datadog-agent/conf.d/redis_sentinel.d/conf.yaml
      Total Runs: 33
      Metric Samples: Last Run: 3, Total: 99
      Events: Last Run: 0, Total: 0
      Service Checks: Last Run: 2, Total: 66
      Average Execution Time : 1ms
      Last Execution Date : 2022-10-31 10:00:54 CET / 2022-10-31 09:00:54 UTC (1667206854000)
      Last Successful Execution Date : 2022-10-31 10:00:54 CET / 2022-10-31 09:00:54 UTC (1667206854000)

      Warning: Error collecting metrics for master jd-redis-production-1: unknown subcommand 'SLAVES'. Try SENTINEL HELP.
```

**Additional environment details (Operating System, Cloud provider, etc):**
Ubuntu 20.04.5 LTS, dedicated server, Redis 7.0.0 + redis stack installed

**Steps to reproduce the issue:**
1. Install redis_sentinel integration, version 1.1.0
2. Configure it for your redis sentinel
3. It fails to send any data due to removed slaves command

**Describe the results you received:**
No data sent to datadog

**Describe the results you expected:**
Integration data sent to datadog

**Additional information you deem important (e.g. issue happens only occasionally):**

Redis changed commands from SLAVES to REPLICAS:
https://redis.io/docs/manual/sentinel/#sentinel-commands

FlorentClarret:
Hi, thanks for opening this issue. Pinging @krasnoukhov as maintainer of the Redis Sentinel integration.

krasnoukhov:
Unfortunately we don't use this integration any longer (moved to ElastiCache). Any chance someone else would be interested in taking over?

FlorentClarret:
> Unfortunately we don't use this integration any longer (moved to ElastiCache). Any chance someone else would be interested in taking over?

I understand. I opened [a PR](https://github.com/DataDog/integrations-extras/pull/1589) to update the codeowners accordingly. 

@Akkowicz We will _try_ to take a look at the issue but I can't tell you when exactly

PR #3071, opened 2026-07-23
redis_sentinel: Add SSL/TLS, username/ACL, and Redis 7+ replica support

### LINKED ISSUE #2938: redis_sentinel: No SSL/TLS support for Sentinel connections
## Problem

In [`redis_sentinel.py` line 33](https://github.com/DataDog/integrations-extras/blob/master/redis_sentinel/datadog_checks/redis_sentinel/redis_sentinel.py#L33), the Redis connection is created without any SSL/TLS parameters:

```python
redis_conn = redis.StrictRedis(host=host, port=port, password=password, db=0)
```

This makes it impossible to connect to Redis Sentinel instances that require TLS, which is increasingly common in production environments (e.g. AWS ElastiCache, Azure Cache for Redis, or any hardened deployment).

## Expected Behavior

The check should support `ssl`, `ssl_certfile`, `ssl_keyfile`, `ssl_ca_certs`, and related parameters from the instance config, similar to the core [Redis integration](https://docs.datadoghq.com/integrations/redisdb/#configuration-options).

## Suggested Fix

Pass SSL parameters from the instance config through to `redis.StrictRedis`:

```python
ssl = instance.get('ssl', False)
ssl_kwargs = {}
if ssl:
    ssl_kwargs = {
        'ssl': True,
        'ssl_certfile': instance.get('ssl_certfile'),
        'ssl_keyfile': instance.get('ssl_keyfile'),
        'ssl_ca_certs': instance.get('ssl_ca_certs'),
        'ssl_cert_reqs': instance.get('ssl_cert_reqs', 'required'),
    }

redis_conn = redis.StrictRedis(host=host, port=port, password=password, db=0, **ssl_kwargs)
```

ian28223:
## Summary

Modernizes the Redis Sentinel check to support TLS-enabled deployments and Redis 6+ ACLs.

Supersedes #2939 (opened by @amendez-primer, auto-closed for inactivity after CI kept failing). Same change set, with the CI failures fixed and review feedback addressed:

- **Formatting**: ran `ddev test --fmt redis_sentinel` with the current ddev version, which the original review asked for but the PR never got an updated pass on.
- **Dependency validation**: `ddev validate dep` rejects unstable `>=` pins. The `redis` dependency is now pinned exactly (`redis==7.3.0`, matching `redisdb` in `integrations-core`) instead of `redis>=4.5.0`.
- **Minimum base package floor**: bumped `datadog-checks-base` from `>=4.2.0` to `>=37.10.0`. The old floor imports `six.moves`, which isn't declared as a hard dependency, and its `[deps]` extra pins an old `ddtrace` with no `cp313` wheel — both broke the "minimum base package" CI job on unmodified `master` too, unrelated to this PR's changes.
- **`_get_sentinel_replicas()` removed**: a reviewer caught that `redis.StrictRedis` never had a `sentinel_replicas` method in any released version, so the `hasattr()` check was always `False` and this always fell through to `sentinel_slaves()` — dead code with an inaccurate description (see below).
- **`ssl_check_hostname` added**: a reviewer caught that redis-py defaults hostname verification to `True` and only disables it when `ssl_cert_reqs` is `CERT_NONE`, so a sentinel reached by IP with a hostname-mismatched cert had no escape short of disabling certificate verification entirely. Matches `redisdb`'s existing option.

Fixes #2938, #2395

### 1. SSL/TLS support (#2938)
The check previously created `redis.StrictRedis` connections without any SSL parameters, making it impossible to connect to TLS-enabled Sentinel instances (e.g. AWS ElastiCache, Azure Cache for Redis).

Added support for: `ssl`, `ssl_certfile`, `ssl_keyfile`, `ssl_ca_certs`, `ssl_cert_reqs`, `ssl_check_hostname` — matching the core `redisdb` integration.

### 2. Username / ACL support (#2395)
Redis 6+ introduced ACLs requiring a `username` parameter. Without it, `AUTH` fails with: `AUTH <password> called without any password configured for the default user`.

Added `sentinel_username` instance config option, passed through to `StrictRedis(username=...)`.

### 3. Socket timeout
Added `socket_timeout` config option (default: 5s) to prevent the check from hanging indefinitely on unreachable sentinels.

Note: an earlier revision of this PR also claimed a Redis 7+ `SENTINEL REPLICAS` fallback (#1586). That code path was dead (see above) and the underlying premise didn't hold up either — `SENTINEL SLAVES` is still a live alias for `SENTINEL REPLICAS` in current Redis, not a removed subcommand — so it's been dropped rather than reimplemented.

## Files changed

| File | Change |
|------|--------|
| `redis_sentinel.py` | SSL, username, timeout params in `_load_config()` |
| `conf.yaml.example` | Documented all new config options |
| `test_redis_sentinel.py` | Tests for SSL, username, timeout config |
| `pyproject.toml` | Bumped `redis` dep to `==7.3.0` (pinned, matches `redisdb`); bumped `datadog-checks-base` floor to `>=37.10.0` |
| `__about__.py` / `CHANGELOG.md` | Bumped to 1.2.0 |

## Backward compatibility
- All new options are optional with sensible defaults
- Without `ssl: true`, behavior is identical to before

## Test plan
- [x] `ddev test --lint redis_sentinel` passes (ruff format + check)
- [x] `ddev test redis_sentinel -- -m unit` passes (4/4 unit tests)
- [x] CI: lint, unit/integration tests, and minimum-base-package job all green
- [ ] Integration tests (`ddev test redis_sentinel`) — pass locally except for an arm64/amd64 emulation issue with the `erichsu/redis-sentinel` test image unrelated to this change; CI runs on amd64 and is green

🤖 Generated with [Claude Code](https://claude.com/claude-code)


redis_sentinel/datadog_checks/redis_sentinel/redis_sentinel.py
```diff
@@ -24,13 +24,40 @@ def _load_config(self, instance):
             raise ConfigurationError('Configuration error. "sentinel_port" must be an int.')
 
         passwd = instance.get('sentinel_password', None)
-
-        return host, port, passwd
+        username = instance.get('sentinel_username', None)
+        socket_timeout = instance.get('socket_timeout', 5)
+
+        ssl_kwargs = {}
+        if instance.get('ssl', False):
```
HadhemiDD, line 31: **_Powered by Claude_**
Request: I'd add ssl_check_hostname to the passthrough. redis-py defaults it to True (client.py:244) and only auto-disables it when cert_reqs == CERT_NONE (connection.py:1978-1980), so it applies at both the documented default of 2 and at 1. Sentinel configs commonly carry IPs, so a user setting ssl: true against an IP-addressed or SAN-mismatched sentinel gets SSLCertVerificationError and the only escape offered here is ssl_cert_reqs: 0, which turns off certificate verification wholesale. That's a big hammer for a hostname mismatch, and it makes the headline feature unusable-or-insecure for a common topology. redisdb — the integration this PR says it matches — includes ssl_check_hostname in list_params (redisdb.py:91) and documents it for exactly this case: "Set to false if your certificate does not match the connection hostname (for example, if connecting by IP)" (redisdb/assets/configuration/spec.yaml:121-128). It needs a matching conf.yaml.example entry too.

redis_sentinel/pyproject.toml
```diff
@@ -29,15 +29,15 @@ classifiers = [
...
+    "datadog-checks-base>=37.10.0",
 ]
 dynamic = [
     "version",
 ]
 
 [project.optional-dependencies]
 deps = [
-    "redis==2.10.5",
+    "redis==7.3.0",
```
HadhemiDD, line 40: Request: There's no version bump and no CHANGELOG entry, so none of this can actually ship. __about__.py is still 1.1.1 and CHANGELOG.md still ends at ## 1.1.1 / 2023-07-26. Extras uses a hand-written CHANGELOG.md (no towncrier fragment dirs anywhere in the repo), and recent merges all touch both files — e8255de5 (redis_enterprise_prometheus #3078), 2a234ce3 (grpc_check #2991), 33d76282 (ping #3081). I'd bump to 1.2.0 and add an ***Added***: section referencing [#3071].

redis_sentinel/datadog_checks/redis_sentinel/redis_sentinel.py
```diff
@@ -44,6 +71,15 @@ def _process_instance_master(self, redis_conn, master_name, base_tags):
         self._process_slaves_stats(redis_conn, master_name, base_tags, master_tags)
         self._process_sentinels_stats(redis_conn, master_name, base_tags, master_tags)
 
+    def _get_sentinel_replicas(self, redis_conn, master_name):
```
HadhemiDD: Request: This helper never takes its primary branch, so fix #3 isn't delivered. redis-py 7.3.0's redis/commands/sentinel.py defines sentinel_slaves (line 66) and no sentinel_replicas; the same is true on master. Redis has no `__getattr__`, so hasattr(redis_conn, 'sentinel_replicas') is always False and every call falls straight through to sentinel_slaves() — identical to the code on master. The PR description's "(redis-py 4.2+)" isn't accurate.

The issue it targets (#1586) also isn't reproducible as described: `src/sentinel.c` still has `!strcasecmp(c->argv[1]->ptr,"slaves") || !strcasecmp(c->argv[1]->ptr,"replicas")` in Redis 7.4, 8.0, 8.2 and unstable, so SENTINEL SLAVES is a live alias, not a removed subcommand.

I'd delete the helper, restore the direct redis_conn.sentinel_slaves(master_name) call at line 163, and drop claim #3 from the description. If you'd rather actually emit REPLICAS, note that execute_command("SENTINEL REPLICAS", ...) won't work as a drop-in: redis-py registers parse_sentinel_slaves_and_sentinels for "SENTINEL SLAVES" only (`_parsers/helpers.py:890`, and no "SENTINEL REPLICAS" key in the RESP3 map either), so you'd get raw nested arrays and the stats['is_odown'] / stats['master-link-status'] accesses downstream would fail.

If you do keep it, please narrow the handler — except Exception: pass with no logging would swallow AuthenticationError (which subclasses ConnectionError) and NoPermissionError, both newly reachable now that ACLs are supported, then silently re-issue a second command that fails the same way:

```
try:
    return redis_conn.sentinel_replicas(master_name)
except redis.exceptions.NoPermissionError:
    raise
except redis.exceptions.ResponseError as e:
    self.log.debug('SENTINEL REPLICAS unsupported, falling back to SLAVES: %s', e)
```

redis_sentinel/tests/test_redis_sentinel.py
```diff
@@ -92,7 +164,7 @@ def test_down_slaves(aggregator, instance):
     for _ in range(7):
         sentinel_slaves.append({'is_odown': False, 'is_sdown': True})
 
-    with mock.patch('redis.StrictRedis.sentinel_slaves', return_value=sentinel_slaves):
+    with mock.patch.object(check, '_get_sentinel_replicas', return_value=sentinel_slaves):
```
HadhemiDD: Request: I'd revert this to patching the library boundary. The previous mock.patch('redis.StrictRedis.sentinel_slaves', ...) exercised the real path _process_slaves_stats → _get_sentinel_replicas → sentinel_slaves; patching the check's own private method stubs out the new code entirely, so the hasattr branch, the except swallow, and the fallback have zero coverage anywhere in the suite. Nothing forced the change — sentinel_slaves still exists in 7.3.0. mock.patch('redis.StrictRedis.sentinel_slaves', return_value=sentinel_slaves) still works, and if the helper is deleted per my first comment this reverts naturally.

HadhemiDD (changes requested):

ian28223:
Thanks for the thorough review — I verified all four points against redis-py 7.3.0 source directly and they all held up:

1. **`ssl_check_hostname`** — confirmed `SSLConnection.__init__` sets `self.check_hostname = ssl_check_hostname if self.cert_reqs != ssl.CERT_NONE else False`, so it stays `True` by default at `CERT_REQUIRED`/`CERT_OPTIONAL`. Added the passthrough + `conf.yaml.example` entry, matching `redisdb`.
2. **Version bump / CHANGELOG** — bumped to 1.2.0 and added the changelog entry.
3. **`_get_sentinel_replicas()`** — confirmed `hasattr(redis.StrictRedis, 'sentinel_replicas')` is `False` on 7.3.0 (no such method anywhere in `redis.commands.sentinel.SentinelCommands`), so it was dead code. Removed the helper, restored the direct `sentinel_slaves()` call, and dropped the inaccurate Redis 7+/#1586 claim from the description.
4. **Test mock** — reverted to `mock.patch('redis.StrictRedis.sentinel_slaves', ...)` now that the helper's gone.

Pushed in f8c911b0. Lint and unit tests pass locally.

redis_sentinel/tests/test_redis_sentinel.py
```diff
@@ -46,18 +46,92 @@ def test_load_config():
...
+
+    # SSL enabled with minimal options
+    _, _, _, _, _, ssl_kwargs = c._load_config(
+        {'sentinel_host': 'localhost', 'sentinel_port': 123, 'masters': 'mymaster', 'ssl': True}
+    )
+    assert ssl_kwargs == {'ssl': True}
+
+
+@pytest.mark.unit
+def test_load_config_socket_timeout():
```
HadhemiDD, line 128: **Request**: I'd like a test that the new options actually reach the client. Everything this PR adds is consumed at `redis_sentinel.py:54-63`, and no test touches that path — all four unit tests stop at `_load_config`'s return value, and the only test that calls `check()` is `@pytest.mark.integration` + `dd_environment`, which you note doesn't run locally. Concretely: delete `**ssl_kwargs` from the constructor, or swap `username=username` for `password=username`, and all 4 unit tests still pass green while TLS is silently off. A typo'd kwarg (`ssl_ca_cert`, `ssl_certsfile`) would only surface at customer runtime. This is cheap to close:

```python
@pytest.mark.unit
def test_check_forwards_connection_kwargs():
    c = RedisSentinelCheck(CHECK_NAME, {}, {})
    inst = {'sentinel_host': 'h', 'sentinel_port': 26379, 'masters': [],
            'sentinel_username': 'u', 'ssl': True, 'ssl_ca_certs': '/ca.pem'}
    with mock.patch('redis.StrictRedis') as m:
        c.check(inst)
    kw = m.call_args.kwargs
    assert kw['username'] == 'u'
    assert kw['ssl'] is True and kw['ssl_ca_certs'] == '/ca.pem'
    assert kw['socket_timeout'] == 5
```

redis_sentinel/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # CHANGELOG - Redis Sentinel
 
+## 1.2.0 / 2026-07-31
+
+***Added***:
```
HadhemiDD, line 10: **Request**: The 1.2.0 entry never mentions the dependency changes, which are the highest-impact part of this PR. `redis` jumps five majors, and `datadog-checks-base>=37.10.0` corresponds to **Agent 7.66.0** (`integrations-core/datadog_checks_base/CHANGELOG.md:375` — "## 37.10.0 / 2025-04-17 / Agent 7.66.0"). Anyone on an older Agent running `datadog-agent integration install -t datadog-redis_sentinel==1.2.0` gets a pip resolution failure with nothing in the changelog explaining why. I'd add a `***Changed***` section — the repo renders multiple sections per release (see `redis_enterprise_prometheus/CHANGELOG.md:3-13`), and base-bump changelog lines have direct precedent in `purefa/CHANGELOG.md:34` and `gatekeeper/CHANGELOG.md:19`.

HadhemiDD (dismissed):
Hi @ian28223, this review was generated with Claude Code and has been looked over and approved before posting. I believe the comments below are valid — feel free to discuss any you disagree with.

<details>
<summary><b>Comment legend</b></summary>

**praise**: no action needed, just celebrate!
**note**: informational, no action required.
**question**: seeking clarification or understanding your approach.
**nit**: minor, non-blocking (style, typo). Feel free to ignore.
**suggestion**: optional improvement, recommended but not required.
**request**: a change I believe is necessary before merging.

</details>

redis_sentinel/tests/test_redis_sentinel.py
ian28223: Added `test_check_forwards_connection_kwargs`, using close to your suggested version (mocking `redis.StrictRedis` and asserting on `call_args.kwargs`). Verified it actually catches the gap: locally mutated `username=username` to `password=username` in `check()` and confirmed the new test fails while the other 4 unit tests still pass green — same failure mode you described.

redis_sentinel/CHANGELOG.md
ian28223: Added a `***Changed***` section calling out both bumps, including the Agent 7.66.0+ requirement from the `datadog-checks-base>=37.10.0` floor. Formatted as two separate terse bullets to match the convention in `integrations-core` (e.g. `dcgm/CHANGELOG.md`: "Bump the minimum base check version ([#PR])") rather than one combined sentence.

steveny91 (dismissed):

redis_sentinel/CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+
+* Add SSL/TLS support (`ssl`, `ssl_certfile`, `ssl_keyfile`, `ssl_ca_certs`, `ssl_cert_reqs`, `ssl_check_hostname`) for connecting to TLS-enabled Sentinel instances ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
+* Add `sentinel_username` for Redis 6+ ACL authentication ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
+* Add `socket_timeout` config option to prevent the check from hanging on unreachable sentinels ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
+
+***Changed***:
+
+* Bump `redis` dependency from `2.10.5` to `7.3.0` ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
+* Bump minimum `datadog-checks-base` version to `37.10.0`; this check now requires Agent 7.66.0 or later ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
```
steveny91:
```suggestion
## 1.2.0 / 2026-07-31

***Changed***:

* Bump `redis` dependency from `2.10.5` to `7.3.0` ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
* Bump minimum `datadog-checks-base` version to `37.10.0`; this check now requires Agent 7.66.0 or later ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))

***Added***:

* Add SSL/TLS support (`ssl`, `ssl_certfile`, `ssl_keyfile`, `ssl_ca_certs`, `ssl_cert_reqs`, `ssl_check_hostname`) for connecting to TLS-enabled Sentinel instances ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
* Add `sentinel_username` for Redis 6+ ACL authentication ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
* Add `socket_timeout` config option to prevent the check from hanging on unreachable sentinels ([#3071](https://github.com/DataDog/integrations-extras/pull/3071))
```

redis_sentinel/datadog_checks/redis_sentinel/__about__.py
```diff
@@ -1 +1 @@
-__version__ = '1.1.1'
+__version__ = '1.2.0'
```
steveny91: I think changed is a major version bump

redis_sentinel/CHANGELOG.md
ian28223: Applied — reordered so `***Changed***` leads.

redis_sentinel/datadog_checks/redis_sentinel/__about__.py
ian28223: bumped to 2.0.0. The datadog-checks-base>=37.10.0 floor raises the effective Agent requirement to 7.66.0+, which breaks older Agents, so this should be major per semver rather than minor.

steveny91 (approved):

PR #3075, opened 2026-07-23
Fix Ping check on non-English Windows

ddog-nasirthomas:
### What does this PR do?
Fixes the Ping check on non-English Windows. Windows prints ping's output in the system language, and the check couldn't handle that:

- On German (and other non-English) output it crashed with `'utf-8' codec can't decode byte 0x81`.
- Even when it didn't crash, it looked for the English word `time` to read the response time, which Windows translates (e.g. German `Zeit`) — so it reported reachable hosts as down.

The check now reads the output as UTF-8 and finds the response time by its `ms` unit instead of the English word, so it works in any language. 

### Motivation

Customers were seeing this error ``'utf-8' codec can't decode byte 0x81`` for customers that were utilizing non-english hosts.

### Review checklist

- [ ] Git history is clean

### Additional Notes

### DIFF
--- a/ping/datadog_checks/ping/ping.py
+++ b/ping/datadog_checks/ping/ping.py
@@ -44,7 +44,7 @@ class PingCheck(AgentCheck):
             cmd = "ping"
 
         if platform.system() == "Windows":  # pragma: nocover
-            precmd = ["cmd", "/c", "chcp 437 &"]  # Set code page to English for non-US Windows
+            precmd = ["cmd", "/c", "chcp 65001 &"]  # UTF-8 code page so non-ASCII ping output decodes correctly
             countOption = "-n"
             timeoutOption = "-w"
             # The timeout option is in ms on Windows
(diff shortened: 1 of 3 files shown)


ping/datadog_checks/ping/ping.py
```diff
@@ -85,12 +85,13 @@ def check(self, instance):
 
         try:
             lines = self._exec_ping(timeout, host)
-            regex = re.compile(r"time[<=]((\d|\.)*)")
+            # Match on the untranslated "ms" unit, since ping localizes the "time" label (e.g. German "Zeit=")
+            regex = re.compile(r"[<=]\s*([\d.]+)\s*ms")
```
dkirov-dd: **request**
I'd keep this only as a fallback, keeping the old regex as the primary one.
Unless we are absolutely certain that there is no way another time unit is returned by the command (e.g. `s`, `µs`, etc.)

dkirov-dd (changes requested):
Looks good but I have one blocker

dkirov-dd (approved):

PR #3092, opened 2026-08-04
Fix Ping check on non-English Windows by decoding with the OEM code page

ddog-nasirthomas:
### What does this PR do?

Fixes the `ping` check crashing on non-English Windows. On Windows the check now runs `ping` via `subprocess.run` and decodes its output with the **OEM code page** (`decode("oem", errors="replace")`) instead of going through `get_subprocess_output`, which hardcodes a UTF-8 decode. The `chcp 65001` approach from 1.0.3 is removed. 

### Motivation

1.0.3 tried to fix this by switching the console to UTF-8 (`chcp 65001`), but that turned out not to change the encoding ping actually uses , so a customer on German Windows still hit the same crash (`'utf-8' codec can't decode byte 0x81`) after updating.

### Review checklist

- [ ] Git history is clean

### Additional Notes

The code fix was able to be tested on a customer's non-English windows host and we confirmed it resolved the crash.

### DIFF
--- a/ping/CHANGELOG.md
+++ b/ping/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG - Ping
 
+## 1.0.4 / 2026-08-04
+
+***Fixed***:
+
+* Fix the Ping check still crashing on non-English Windows. The ``chcp 65001`` approach in 1.0.3 did not change the encoding of ping's output, so localized output still failed with ``'utf-8' codec can't decode byte 0x81``. On Windows the check now decodes ping's output using the OEM code page instead of UTF-8 ([#<PR>](https://github.com/DataDog/integrations-extras/pull/<PR>)).
+
 ## 1.0.3 / 2026-07-23
 
 ***Fixed***:
--- a/ping/CHANGELOG.md
+++ b/ping/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ***Fixed***:
 
-* Fix the Ping check still crashing on non-English Windows. The ``chcp 65001`` approach in 1.0.3 did not change the encoding of ping's output, so localized output still failed with ``'utf-8' codec can't decode byte 0x81``. On Windows the check now decodes ping's output using the OEM code page instead of UTF-8 ([#<PR>](https://github.com/DataDog/integrations-extras/pull/<PR>)).
+* Fix the Ping check still crashing on non-English Windows. The ``chcp 65001`` approach in 1.0.3 did not change the encoding of ping's output, so localized output still failed with ``'utf-8' codec can't decode byte 0x81``. On Windows the check now decodes ping's output using the OEM code page instead of UTF-8 ([#3092](https://github.com/DataDog/integrations-extras/pull/3092)).
 
 ## 1.0.3 / 2026-07-23
 
--- a/ping/datadog_checks/ping/ping.py
+++ b/ping/datadog_checks/ping/ping.py
@@ -6,11 +6,12 @@ import subprocess
 
 from datadog_checks.base import AgentCheck, ConfigurationError
 from datadog_checks.base.errors import CheckException
-from datadog_checks.base.utils.subprocess_output import SubprocessOutputEmptyError, get_subprocess_output
+from datadog_checks.base.utils.subprocess_output import get_subprocess_output
 
 
 class PingCheck(AgentCheck):
     SERVICE_CHECK_NAME = "network.ping.can_connect"
+    WINDOWS_OUTPUT_ENCODING = "oem"
 
     def _load_conf(self, instance):
         # Fetches the conf
@@ -47,6 +48,7 @@ class PingCheck(AgentCheck):
             timeoutOption = "-w"
             # The timeout option is in ms on Windows
             # https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/ping
+            subprocess_timeout = timeout + 10
             timeout = timeout * 1000
         elif platform.system() == "Darwin":
             countOption = "-c"
@@ -67,7 +69,7 @@ class PingCheck(AgentCheck):
         self.log.debug("Running: %s", " ".join(command))
 
         if platform.system() == "Windows":  # pragma: nocover
-            lines, err, retcode = self._exec_ping_windows(command)
+            lines, err, retcode = self._exec_ping_windows(command, subprocess_timeout)
         else:
             lines, err, retcode = get_subprocess_output(command, self.log, raise_on_empty_output=True)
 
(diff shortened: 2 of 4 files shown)


ping/datadog_checks/ping/ping.py
```diff
@@ -65,19 +63,29 @@ def _exec_ping(self, timeout, target_host):
...
-        )
         self.log.debug("ping returned %s - %s - %s", retcode, lines, err)
         if retcode != 0:
             raise CheckException("ping returned {}: {}".format(retcode, err))
 
         return lines
 
+    def _exec_ping_windows(self, command):  # pragma: nocover
+        # ping emits localized output in the OEM code page; get_subprocess_output would hardcode a UTF-8 decode.
+        result = subprocess.run(command, capture_output=True)
```
alexeypilyugin: Dropping the `cmd /c` wrapper means subprocess.run no longer goes through a shell, so failures that used to come back as a non-zero exit code (caught by retcode != 0 at line 75 → CheckException → DOWN) can now raise Python exceptions that check()'s except CheckException at line 106 doesn't catch. Concretely: an 8-block IPv6 host builds cmd = "ping6" (line 41), and Windows ships no ping6.exe, so subprocess.run(["ping6", ...]) raises FileNotFoundError — an unhandled traceback instead of a CRITICAL service check. While you're touching this call, it's also worth adding timeout= (currently only -w bounds ping's own wait, not the subprocess itself) so a wedged process can't block the collector.

ping/datadog_checks/ping/ping.py
```diff
@@ -65,19 +63,29 @@ def _exec_ping(self, timeout, target_host):
...
-            self.log,
-            raise_on_empty_output=True,
-        )
         self.log.debug("ping returned %s - %s - %s", retcode, lines, err)
         if retcode != 0:
             raise CheckException("ping returned {}: {}".format(retcode, err))
 
         return lines
 
+    def _exec_ping_windows(self, command):  # pragma: nocover
```
alexeypilyugin: This method is the entire fix for a bug that has already escaped once (the 1.0.3 chcp 65001 attempt), yet it's # pragma: nocover with no tests, and test-all-windows.yml is jobs: {} — nothing in CI exercises this path. The "oem" codec itself can't be tested off Windows (codecs.lookup("oem") raises LookupError), but the branching logic (empty-output raise, retcode/err propagation) can be, if the codec is injectable so a test can substitute a real single-byte codec like cp850 and feed it the exact 0x81 byte from the reported crash. Worth confirming in the PR description too: was this validated against a real non-English Windows host? GetOEMCP() can diverge from GetConsoleOutputCP() — the same class of encoding-API mismatch that made the 1.0.3 fix ineffective.

ping/datadog_checks/ping/ping.py
```diff
@@ -65,19 +63,29 @@ def _exec_ping(self, timeout, target_host):
...
+        # ping emits localized output in the OEM code page; get_subprocess_output would hardcode a UTF-8 decode.
+        result = subprocess.run(command, capture_output=True)
+        out = result.stdout.decode("oem", errors="replace")
+        err = result.stderr.decode("oem", errors="replace")
+        if not out:
+            raise SubprocessOutputEmptyError("get_subprocess_output expected output but had none.")
```
alexeypilyugin: The error message here still says "get_subprocess_output expected output but had none.", but this method exists specifically to bypass get_subprocess_output. Since SubprocessOutputEmptyError isn't caught by check()'s except CheckException, this message is exactly what surfaces in logs when this path fails, and it will misdirect debugging.

ping/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG - Ping
 
+## 1.0.4 / 2026-08-04
+
+***Fixed***:
+
+* Fix the Ping check still crashing on non-English Windows. The ``chcp 65001`` approach in 1.0.3 did not change the encoding of ping's output, so localized output still failed with ``'utf-8' codec can't decode byte 0x81``. On Windows the check now decodes ping's output using the OEM code page instead of UTF-8 ([#<PR>](https://github.com/DataDog/integrations-extras/pull/<PR>)).
```
alexeypilyugin: let's add the PR number to the link

ping/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG - Ping
```
alexeypilyugin, line 1: probably worth updating the README as well, as it says "Versions prior to **1.0.3** may not work correctly if the installed Windows language is not set to English."

alexeypilyugin (changes requested):

ping/datadog_checks/ping/ping.py
ddog-nasirthomas: The customer tested a similar fix on their own machine.  It was a similar decode-based fix on their own German Windows host and confirmed it resolves the crash. Their change is slightly different, but it is a similar approach decoding ping's output.

I'll update the PR description since I wasn't able to verify this exact fix on a non-English Windows host, but if needed we could possibly give the customer the file to test out. I could also possibly get a sandbox of this up but I think I would need an non-english EC2 instance or an Azure vm

domalessi (approved):

ping/datadog_checks/ping/ping.py
```diff
@@ -65,19 +65,34 @@ def _exec_ping(self, timeout, target_host):
             countOption = "-c"
             timeoutOption = "-W"
 
-        self.log.debug("Running: %s %s %s %s %s %s", cmd, countOption, "1", timeoutOption, timeout, target_host)
+        command = [cmd, countOption, "1", timeoutOption, str(timeout), target_host]
+        self.log.debug("Running: %s", " ".join(command))
+
+        if platform.system() == "Windows":  # pragma: nocover
```
alexeypilyugin: `# pragma: nocover` can now be removed

ping/tests/test_ping.py
```diff
@@ -68,6 +83,34 @@ def test_localized_output(aggregator, instance_response_time):
     aggregator.assert_all_metrics_covered()
 
 
+def test_windows_oem_decode(aggregator, instance_response_time):
```
alexeypilyugin: Looks like this test won't ever fail, because the regex in the check is [searching](https://github.com/DataDog/integrations-extras/blob/5e5f38cbda5aae7a1e4f5eecbfe555b48892ae06/ping/datadog_checks/ping/ping.py#L103C34-L103C39) for english characters and the fallback [searches](https://github.com/DataDog/integrations-extras/blob/5e5f38cbda5aae7a1e4f5eecbfe555b48892ae06/ping/datadog_checks/ping/ping.py#L108) for a number followed by english "ms". 

WINDOWS_GERMAN_OUTPUT will go through the fallback which will only see "=3ms", and they're ascii. 

Instead, smth like this could work

```
def test_windows_oem_decode(self):
    check = PingCheck("ping", {}, {})
    check.WINDOWS_OUTPUT_ENCODING = "cp850"
    proc = mock.Mock(stdout=WINDOWS_GERMAN_OUTPUT, stderr=b"", returncode=0)

    with mock.patch.object(subprocess, "run", return_value=proc):
        lines, err, retcode = check._exec_ping_windows(["ping", "-n", "1", "127.0.0.1"], 14)

    # This is the actual regression check: decode with the wrong codec
    # and "ausgeführt" comes back as mojibake/replacement chars, not this string.
    assert "ausgeführt" in lines
    assert "\ufffd" not in lines
```

alexeypilyugin (changes requested):

alexeypilyugin (approved):

alexeypilyugin:
@jstanton617 would you mind looking at this PR?

ddog-nasirthomas:
I thought integrations-extras doesn't use devflow/merge unless something has changed recently.

jstanton617 (commented):
LGTM

jstanton617 (approved):

PR #3108, opened 2026-08-11
Add RUM dataflows to the RUM SDK tiles

philjlee:
**Jira:** [TXP-277](https://datadoghq.atlassian.net/browse/TXP-277)

Adds `assets/dataflows.yaml` to the 15 RUM SDK tiles, so each one declares the RUM events its SDK sends to Datadog.

### What each file says

```yaml
provides:
  - id: rum-vue-events
    always_on: true
    granular: false
    data_type: rum_events
    direction: inbound
```

- **`always_on: true`** — the tile exists to describe an installed SDK. Once it's in the page or app it sends RUM events; there's no per-tile toggle to turn that off.
- **`granular: false`** — granular is reserved for product-specific onboarding (`llm_completions`, `cloud_cost`, `mcp_tool_calls`).
- **`direction: inbound`** — data flows into Datadog.
- **One dataflow per tile.** Splitting is for logically distinct data with different toggles; a RUM SDK emits one stream of events.

### Dataflow ids

Ids come from the **manifest `app_id`**, not the directory name — those differ for two tiles: `rum_nextjs` → `rum-next-plugin`, `rum_nuxt` → `rum-nuxt-plugin`. Getting this wrong would have produced ids that don't correspond to any app.

| Tile | Dataflow id |
| --- | --- |
| `rum_android` | `rum-android-events` |
| `rum_angular` | `rum-angular-events` |
| `rum_cypress` | `rum-cypress-events` |
| `rum_expo` | `rum-expo-events` |
| `rum_flutter` | `rum-flutter-events` |
| `rum_ios` | `rum-ios-events` |
| `rum_javascript` | `rum-javascript-events` |
| `rum_nextjs` | `rum-next-plugin-events` |
| `rum_nuxt` | `rum-nuxt-plugin-events` |
| `rum_react` | `rum-react-events` |
| `rum_react_native` | `rum-react-native-events` |
| `rum_roku` | `rum-roku-events` |
| `rum_salesforce_lwc` | `rum-salesforce-lwc-events` |
| `rum_shopify` | `rum-shopify-events` |
| `rum_vue` | `rum-vue-events` |

The documented convention is `{app-id}-{data-type}`, which would give `rum-vue-rum-events`. Since every app_id here already begins with `rum-`, the second one is dropped: the id still ends in the data noun and stays unique. Dataflow ids are globally unique and effectively permanent — the validator refuses to remove a dataflow that another app `uses` — so they're worth getting right the first time.

### Why 15 tiles and not 18

Three tiles match "rum" but do not provide RUM events, and are deliberately excluded: **`flagsmith-rum`**, **`split-rum`**, **`statsig_rum`**.

All three are feature-flag integrations that *enrich* RUM. Their setup instructions have the vendor SDK call `datadogRum.addFeatureFlagEvaluation(...)` — the RUM Browser SDK still does the sending, so the provider of that data is `rum_javascript`, not them. Their manifests agree: they're categorized `Configuration & Deployment` / `Developer Tools`, while the 15 included tiles carry `Metrics` / `Tracing` / `Mobile`. Declaring `provides: rum_events` on them would assert something untrue.

If we want to model the enrichment relationship, `uses` is the right mechanism and a separate change.

### Validation

`ddev` does not validate `dataflows.yaml` at all — the only enforcement is server-side in catalogassets. So rather than trust the schema by eye, I ran all 15 payloads through the **real validator** (`unmarshalDataflowsYAML` + `config.Validate()` from `domains/integrationscatalog/libs/catalogassetslib`) in a throwaway test, asserting for every file: it parses, exactly one `provides` and no `uses`, `data_type == rum_events`, `always_on` true, `granular` false, `direction` inbound, and all 15 ids distinct.

```
=== RUN   TestRumTilesRealFixtures
--- PASS: TestRumTilesRealFixtures (0.00s)
```

That test was temporary and is not part of either PR.

Two further checks:

- **`ddev --here validate all`** output is byte-for-byte identical with and without these 15 files (7 pre-existing errors, 152 files valid, all in unrelated checks) — confirming this change introduces no validation regression, and confirming `ddev` ignores the asset.
- **No id collisions.** No dataflow id beginning with `rum-` exists anywhere in `integrations-extras`, `integrations-core`, or `integrations-internal-core`. Caveat: the core and internal-core clones I scanned are local, so a very recently merged `rum-*` id would not have shown up.

[TXP-277]: https://datadoghq.atlassian.net/browse/TXP-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rum_android/assets/dataflows.yaml
+++ b/rum_android/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-android-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_angular/assets/dataflows.yaml
+++ b/rum_angular/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-angular-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_cypress/assets/dataflows.yaml
+++ b/rum_cypress/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-cypress-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_expo/assets/dataflows.yaml
+++ b/rum_expo/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-expo-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_flutter/assets/dataflows.yaml
+++ b/rum_flutter/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-flutter-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_ios/assets/dataflows.yaml
+++ b/rum_ios/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-ios-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_javascript/assets/dataflows.yaml
+++ b/rum_javascript/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-javascript-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_nextjs/assets/dataflows.yaml
+++ b/rum_nextjs/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-next-plugin-events
+    always_on: true
+    granular: false
+    data_type: rum_events
+    direction: inbound
--- a/rum_nuxt/assets/dataflows.yaml
+++ b/rum_nuxt/assets/dataflows.yaml
@@ -1,0 +1,6 @@
+provides:
+  - id: rum-nuxt-plugin-events
+    always_on: true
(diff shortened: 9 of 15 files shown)


philjlee:
`validate-assets` is currently red. This is deploy lag on the dependency, not a problem with these files.

All 15 annotations report the same single error:

```
invalid Dataflow.DataType: must be one of [metrics logs traces cloud_cost dora_metrics
events llm_completions openlineage security_findings threat_intel_feed resources
mcp_tool_calls recommendations federated_sql_queries], got 'rum_events'
```

That enumerated list is exactly the **pre-merge** 14-value allowlist. [ddoghq/dd-source#50640](https://github.com/ddoghq/dd-source/pull/50640) appends `rum_events` as the 15th and merged at 18:56Z (`df3e967`); this validation ran 18:58–19:06Z. The deployed catalogassets worker had not yet picked up the change, so it was still validating against the old list.

Worth noting what the failure *doesn't* say: `data_type` is the only complaint on any of the 15 files. The ids, `always_on`, `granular`, `direction`, and the `provides`/`uses` structure all passed the real server-side validator.

Re-running once the worker has rolled out `df3e967` should clear this — per the check's own guidance, syncing with the default branch re-triggers it.

justinvos (approved):
Your explanation with the dataflow changes makes sense to me.

I did notice we are missing integration tiles for RUM SDK integrations with:
- Electron
- c++
- Kotlin
- .NET Maui

Not something for this PR but maybe important to follow up on.

PR #3128, opened 2026-08-24
Rename get_default_config to get_default_metrics in redis_cloud/redis_enterprise checks

philjlee:
## What

Renames the `get_default_config()` method on `RedisCloudCheck` and
`RedisEnterpriseCheck` to `get_default_metrics()`.

## Why

Both checks defined their own `get_default_config()` returning a list of
metric groups. A newer `datadog-checks-base` adds its own
`get_default_config()` hook to `OpenMetricsBaseCheckV2`, used internally by
`get_config_with_defaults()` to seed scraper config from the check's
generated `config_models` defaults (expects key/value pairs). Because
`pyproject.toml` pins `datadog-checks-base>=37.20.0` with no upper bound,
CI resolves the latest version, whose base class method gets shadowed by
each check's same-named override — `get_config_with_defaults()` then calls
the check's list-returning method instead and `dict()` raises:

```
ValueError: dictionary update sequence element #0 has length 65; 2 is required
```

Confirmed pre-existing and unrelated to any other open PR: reproduced on an
unmodified checkout of `master` via `uvx --python 3.13 ddev@18.0.0 -x test
redis_cloud` — the failure is byte-for-byte identical. It's only visible in
CI when a PR happens to touch files inside `redis_cloud/` or
`redis_enterprise/`, since `TARGET=changed` scopes the test matrix to
touched directories.

## How

Renamed the check-specific method (and its one call site in
`_parse_config`) in both integrations from `get_default_config` to
`get_default_metrics`, which no longer collides with the base class's
config-defaults hook.

Verified locally with `ddev@18.0.0` (the version CI's `test` workflow
resolves): `redis_cloud` 5/5 passing, `redis_enterprise` 7/7 passing, lint
clean.

### DIFF
--- a/redis_cloud/datadog_checks/redis_cloud/check.py
+++ b/redis_cloud/datadog_checks/redis_cloud/check.py
@@ -16,7 +16,7 @@ class RedisCloudCheck(OpenMetricsBaseCheckV2):
     def _parse_config(self):
         self.scraper_configs = []
         metrics_endpoint = self.instance.get('openmetrics_endpoint')
-        metrics = self.get_default_config()
+        metrics = self.get_default_metrics()
 
         additional = []
         groups = self.instance.get('extra_metrics', [])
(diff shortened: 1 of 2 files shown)


philjlee:
Superseded — combined into #3127. Both PRs touched the same redis_cloud/redis_enterprise directories, and under CI's TARGET=changed scoping neither could pass independently of the other (whichever was checked standalone would fail on the other's pre-existing bug). Folded the get_default_config rename into #3127 alongside the drift fix.

PR #3139, opened 2026-08-27
upsc: Correct two metric names in metadata.csv

pault-pg:
### What does this PR do?

Corrects two metric names in `upsc/metadata.csv`. The check reports `upsc.` followed by the NUT variable name, and the NUT variables are `ups.status` and `ups.load`, so the metrics are `upsc.ups.status` and `upsc.ups.load`. `metadata.csv` listed them as `upsc.status` and `upsc.load`.

### Motivation

Those two metrics are emitted under different names than `metadata.csv` documents, so the units and descriptions there do not reach them.

Verified on a host running the integration:

```
datadog-agent check upsc | grep '"metric"' | sort -u
```

which lists `upsc.ups.status` and `upsc.ups.load`.

### Review checklist

- [x] Git history is clean

### Additional Notes

Metadata only, with no change to the data the check reports, so there is no version bump or changelog entry, matching #2864. The `no-changelog` label is probably needed. No tests, since nothing executable changed.

LLM assistance was used to compare the emitted metric names against `metadata.csv` and to write this description.

### DIFF
--- a/upsc/metadata.csv
+++ b/upsc/metadata.csv
@@ -1,5 +1,5 @@
 metric_name,metric_type,interval,unit_name,per_unit_name,description,orientation,integration,short_name,curated_metric
-upsc.status,gauge,,,,UPS status,1,upsc,status,
+upsc.ups.status,gauge,,,,UPS status,1,upsc,status,
 upsc.battery.charge,gauge,,percent,,UPS battery charge,1,upsc,battery charge %,
 upsc.battery.charge.low,gauge,,percent,,UPS remaining battery level when UPS switches to Low Battery,1,upsc,low charge %,
 upsc.battery.charge.warning,gauge,,percent,,UPS battery level when UPS switches to Warning state,1,upsc,warn charge %,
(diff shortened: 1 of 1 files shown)


platinummonkey (approved):

janine-c (approved):

pault-pg:
The two failing checks are one cause, and it looks pre-existing rather than something this PR introduced.

`test / check` is the alls-green aggregate, so it fails because the other one did. The real failure is `test-minimum-base-package`, which installs the oldest base package this integration declares:

```toml
dependencies = ["datadog-checks-base>=4.2.0"]
```

That version cannot be imported on Python 3.13. Reproducible with nothing from this PR involved:

```console
$ uv venv --python 3.13 /tmp/minbase
$ uv pip install --python /tmp/minbase/bin/python 'datadog-checks-base==4.2.0'
$ /tmp/minbase/bin/python -c "import datadog_checks.base"
ModuleNotFoundError: No module named 'checks'
...
ModuleNotFoundError: No module named 'six'
```

which is the same chain CI reports while collecting `tests/test_upsc.py`. The regular `test` job passes, only the minimum-version variant fails, and this PR only edits `metadata.csv`.

`>=4.2.0` is declared by 17 integrations in this repository, against 26 on `>=37.20.0`, so upsc is not unusual here. The job only runs for an integration when that integration is touched, which is presumably why it does not come up more often.

I am happy to raise the floor for upsc in a separate PR if that would help, but which version to pick looks like a support-policy decision rather than simply the oldest one that imports, so I would rather you name it than guess.

LLM assistance was used to isolate the failing job and to write this comment.

PR #3159, opened 2026-09-15
feat(amazon-vpc-cni): add new integration for Amazon VPC CNI

willianccs:
### What does this PR do?

Adds a new Agent-based integration for the [Amazon VPC CNI plugin for Kubernetes](https://github.com/aws/amazon-vpc-cni-k8s).

The check scrapes the Prometheus metrics exposed by the VPC CNI plugin (`awscni_*`) and maps them to `amazon_vpc_cni.*`, covering IP address allocation, ENI usage, IPAMD activity, AWS API calls, and startup duration.

### Motivation

These metrics are currently collected as `kube_system.awscni_*` custom metrics, incurring custom metric billing. This integration promotes them to an official, curated integration.

### Review checklist

- [x] PR has a meaningful title
- [x] Feature has tests
- [x] Git history is clean

### Additional Notes

Metrics sourced from the VPC CNI plugin's Prometheus endpoint (port 61678), enabled via `ENABLE_PROMETHEUS_METRICS=true`.


buraizu:
Thanks for the PR! I've created DOCS-15717 for documentation team review.

amazon_vpc_cni/datadog_checks/amazon_vpc_cni/metrics.py
```diff
@@ -0,0 +1,33 @@
...
+    'awscni_total_ipv4_prefixes': 'total_ipv4_prefixes',
+    'awscni_assigned_ip_per_cidr': 'assigned_ip_per_cidr',
+    'awscni_no_available_ip_addresses': 'no_available_ip_addresses',
+    'awscni_assigned_ip_per_eni': 'assigned_ip_per_eni',
+    ## Startup duration
+    'awscni_ipamd_startup_duration_seconds': 'ipamd_startup_duration_seconds',
+    'awscni_ipamd_node_initialization_duration_seconds': 'ipamd_node_initialization_duration_seconds',
+    ## Connmark
+    'awscni_connmark_backend': 'connmark_backend',
+    'awscni_connmark_reconcile_total': 'connmark_reconcile_total',
```
lucia-sb: **Request**: Nice catch overall on the metric mapping, but I think there's a subtle bug here with `awscni_connmark_reconcile_total` — it's keyed with the `_total` suffix intact, and `prometheus_client`'s parser actually strips `_total` from counter family names before OpenMetricsV2 looks the name up in `METRIC_MAP` (`prometheus_client/parser.py`'s `build_metric`; `MetricTransformer.get` in `v2/transform.py` does an exact-name match with no fallback). So as written, this key never matches and `amazon_vpc_cni.connmark_reconcile_total.count` — even though it's declared in `metadata.csv` and asserted in `tests/common.py` — never actually gets emitted. Mapping the stripped name should fix it without needing to rename the `metadata.csv` row, e.g. `'awscni_connmark_reconcile': {'name': 'connmark_reconcile_total'}`.

amazon_vpc_cni/tests/test_unit.py
```diff
@@ -0,0 +1,35 @@
...
+
+def test_connect_exception(dd_run_check, aggregator, caplog):
+    with pytest.raises(Exception, match="Failed to resolve 'invalid-hostname'|Max retries exceeded"):
+        check = AmazonVpcCniCheck('amazon_vpc_cni', {}, [BAD_HOSTNAME_INSTANCE])
+        dd_run_check(check)
+
+    aggregator.assert_service_check('amazon_vpc_cni.openmetrics.health', ServiceCheck.CRITICAL)
+
+
+def test_check_mock_amazon_vpc_cni_metrics(dd_run_check, aggregator, check, mock_prometheus_metrics):
```
lucia-sb, line 20: **Request**: Since all 32 metric assertions here use `at_least=0`, they end up being vacuous (`aggregator.assert_metric` reduces to `len(candidates) >= 0`), so the suite would pass even if zero metrics were emitted — that's likely why the mapping bug on `metrics.py:32` slipped through. Bumping to `at_least=1` and adding `check_symmetric_inclusion=True` to `assert_metrics_using_metadata(...)` would make a future unmapped or misspelled metric fail loudly instead of silently.

.github/CODEOWNERS
```diff
@@ -20,11 +20,12 @@ code-coverage.datadog.yml                                  @DataDog/agent-integr
 /alertnow/                               @Uijong-Kim uijong.kim@bespinglobal.com @DataDog/ecosystems-review
 /algorithmia/                            @koverholt support@algorithmia.io @DataDog/ecosystems-review
 /altostra/                               @yevk @altostra/engineering @DataDog/ecosystems-review
-/ambassador/                             @datawire hello@datawire.io @DataDog/ecosystems-review
-/amixr/                                  @iskhakov ildar@amixr.io @DataDog/ecosystems-review
-/apache-apisix/                          @bisakhmondal bisakhmondal00@gmail.com @DataDog/ecosystems-review
-/apollo/                                 @sachindshinde sachin@apollographql.com @DataDog/ecosystems-review
-/appkeeper/                              @numno1 @DataDog/ecosystems-review
+/amazon_vpc_cni/                         @willianccs willianccs@gmail.com
```
lucia-sb: **Request**: Looks like this branch may be a bit behind `master` — the diff reverts `@DataDog/ecosystems-review` off five unrelated entries (`ambassador`, `amixr`, `apache-apisix`, `apollo`, `appkeeper`) that already have it upstream, which explains the `-5` in the diff stat. The new `amazon_vpc_cni` entry could also use `@DataDog/ecosystems-review` added, matching other community-authored entries like `external_secrets`. A quick rebase onto current `master` should clear this up.

amazon_vpc_cni/metadata.csv
```diff
@@ -0,0 +1,33 @@
...
+amazon_vpc_cni.ec2api_req_count.count,count,,,,The number of requests made to EC2 APIs by the CNI.,0,amazon_vpc_cni,ec2 api requests,
+amazon_vpc_cni.ec2api_error_count.count,count,,,,The number of failed EC2 API requests.,-1,amazon_vpc_cni,ec2 api errors,
+amazon_vpc_cni.force_removed_enis.count,count,,,,The number of ENIs force removed while they had assigned pods.,0,amazon_vpc_cni,force removed enis,
+amazon_vpc_cni.force_removed_ips.count,count,,,,The number of IPs force removed while they had assigned pods.,0,amazon_vpc_cni,force removed ips,
+amazon_vpc_cni.no_available_ip_addresses.count,count,,,,The number of pod IP assignments that fail due to no available IP addresses.,-1,amazon_vpc_cni,no available ip addresses,
+amazon_vpc_cni.connmark_reconcile_total.count,count,,,,The number of connmark reconciliations by result.,0,amazon_vpc_cni,connmark reconciles,
+amazon_vpc_cni.aws_api_latency_ms.count,count,,,,Number of AWS API latency observations.,0,amazon_vpc_cni,aws api latency count,
+amazon_vpc_cni.aws_api_latency_ms.sum,count,,,,Sum of AWS API call latencies.,0,amazon_vpc_cni,aws api latency sum,
```
lucia-sb: **Request**: For the `aws_api_latency_ms` rows, `unit_name` is set to `second` on the `.count`/`.bucket` sub-metrics, but those are dimensionless counts rather than durations — the actual millisecond-valued `.sum`/`.quantile` series are the ones missing a `unit_name`. Other integrations like `kuma`, `argo_workflows`, and `azure_iot_edge` leave `unit_name` blank on `.count`/`.bucket` and set it only on the value-bearing sub-metric (`millisecond` is a valid `ddev` unit), which might be a good pattern to follow here too.

amazon_vpc_cni/datadog_checks/amazon_vpc_cni/check.py
```diff
@@ -0,0 +1,21 @@
+from datadog_checks.base import OpenMetricsBaseCheckV2
+
+from .metrics import METRIC_MAP
+
+
+class AmazonVpcCniCheck(OpenMetricsBaseCheckV2):
+    __NAMESPACE__ = 'amazon_vpc_cni'
+    DEFAULT_METRIC_LIMIT = 0
+
+    def __init__(self, name, init_config, instances):
```
lucia-sb: **Suggestion**: `send_distribution_sums_as_monotonic`/`send_distribution_counts_as_monotonic` look like they're carried over from an OpenMetrics V1-style check — they're only read by the V1 mixin/base check, so on this V2-based check they're unused config (and passed as the string `'true'` rather than `True`). The overridden `__init__`/`check()` are also plain pass-throughs at this point, so both could probably be dropped to simplify the class a bit.

amazon_vpc_cni/datadog_checks/amazon_vpc_cni/metrics.py
```diff
@@ -0,0 +1,33 @@
+METRIC_MAP = {
+    ## IPAMD
+    'awscni_ipamd_error_count': 'ipamd_error_count',
+    'awscni_ipamd_action_inprogress': 'ipamd_action_inprogress',
+    'awscni_reconcile_count': 'reconcile_count',
+    'awscni_add_ip_req_count': 'add_ip_req_count',
+    'awscni_del_ip_req_count': 'del_ip_req_count',
+    'awscni_pod_eni_error_count': 'pod_eni_error_count',
+    ## AWS API
```
lucia-sb, line 9: **Suggestion**: Great coverage overall — I noticed upstream registers 29 `awscni_*` metrics and `METRIC_MAP` covers 27 of them. The two not yet mapped are `awscni_sagemakerapi_req_count` and `awscni_sagemakerapi_error_count`. Might be worth adding them for full parity, unless there's a reason to leave them out that could be called out in a comment.

amazon_vpc_cni/tests/test_e2e.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from datadog_checks.base.constants import ServiceCheck
+from datadog_checks.dev.utils import get_metadata_metrics
+
+from .common import EXPECTED_PROMETHEUS_METRICS
+
+
+@pytest.mark.e2e
+def test_e2e_service_check_ok(dd_agent_check, aggregator, instance, mock_prometheus_metrics):
```
lucia-sb: **Suggestion**: Both e2e tests pull in `mock_prometheus_metrics`, which patches `requests.Session.get` in the pytest process — but since `dd_agent_check` runs the check inside the Agent container, the patch doesn't actually reach it there. It works out fine today because the `dd_environment` container happens to serve the same payload, but the fixture itself isn't doing anything at this layer, so it might be worth dropping from the e2e tests to avoid confusion later.

lucia-sb (changes requested):

lucia-sb:
Thanks for the PR! I left a few comments below.

<details>
<summary><b>Comment legend</b></summary>

**request**: a change I believe is necessary before merging.
**suggestion**: optional improvement, recommended but not required.
**question**: seeking clarification or understanding your approach.
**nit**: minor, non-blocking (style, typo). Feel free to ignore.
**praise**: no action needed, just celebrate!

</details>

amazon_vpc_cni/datadog_checks/amazon_vpc_cni/metrics.py
willianccs: Fixed in fa6472d9. The map key is now `awscni_connmark_reconcile` (without the `_total` suffix, which prometheus_client strips), so `connmark_reconcile_total.count` now emits. Verified with the strengthened unit test.

amazon_vpc_cni/tests/test_unit.py
willianccs: Fixed in fa6472d9. Assertions now use `at_least=1` and `assert_metrics_using_metadata(..., check_symmetric_inclusion=True)`.

.github/CODEOWNERS
willianccs: Fixed in fa6472d9. Restored `@DataDog/ecosystems-review` on ambassador/amixr/apache-apisix/apollo/appkeeper and added it to `amazon_vpc_cni`. Note the branch was actually current with master; the five entries were dropped by an earlier merge resolution, not staleness.

amazon_vpc_cni/metadata.csv
willianccs: Fixed in fa6472d9. One correction: the misplaced `second` units were on the `ipamd_*_duration_seconds` `.count`/`.bucket` rows, not `aws_api_latency_ms` — the `aws_api_latency_ms.sum`/`.quantile` were the ones missing a unit. Applied the stated principle across all three duration families: `.count`/`.bucket` left blank, `.sum`/`.quantile` carry `millisecond`/`second`.

amazon_vpc_cni/datadog_checks/amazon_vpc_cni/check.py
willianccs: Done in fa6472d9. Dropped the unused V1-only `send_distribution_*_as_monotonic` flags and the pass-through `__init__`/`check()`.

amazon_vpc_cni/datadog_checks/amazon_vpc_cni/metrics.py
willianccs: Added in fa6472d9. `awscni_sagemakerapi_req_count` and `awscni_sagemakerapi_error_count` are now mapped, with matching `metadata.csv` rows, test expectations, and fixture samples for full parity.

amazon_vpc_cni/tests/test_e2e.py
willianccs: Removed in fa6472d9. Dropped the unused `mock_prometheus_metrics` fixture from both e2e tests.

amazon_vpc_cni/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The [Amazon VPC CNI plugin for Kubernetes][13] is the networking plugin deployed on each Amazon EC2 node in an Amazon EKS cluster. It manages elastic network interfaces (ENIs) and assigns private IPv4 or IPv6 addresses from the VPC to each pod.
+
+This integration scrapes the Prometheus metrics exposed by the VPC CNI plugin (`awscni_*`) and provides visibility into IP address allocation, ENI usage, AWS API activity, and IPAMD health. It lets you replace ad-hoc `awscni_*` custom metrics with an official, curated integration.
+
+## Setup
+
+### Installation
+
+If you are using Agent v6.8+ follow the instructions below to install the Amazon VPC CNI check on your host. See the dedicated Agent guide for [installing community integrations][1] to install checks with the [Agent Manager][2] or in a [Docker environment][4].
```
joepeeples:
```suggestion
If you are using Datadog Agent v6.8+, follow the instructions below to install the Amazon VPC CNI check on your host. See the dedicated Agent guide for [installing community integrations][1] to install checks with the [Agent Manager][2] or in a [Docker environment][4].
```

joepeeples (dismissed):
Approved with small edit suggestion, thanks!

willianccs:
@lucia-sb could you check again please? Thank you!

lucia-sb (approved):

joepeeples (approved):

PR #3166, opened 2026-09-18
fix(external-secrets): correct reconcile_duration unit and add terminal_reconcile_errors metric

willianccs:
## Summary

Fixes identified while validating the External Secrets integration metrics against the custom metrics.

### 1. `reconcile_duration` unit bug
ESO exposes `reconcile_duration` in nanoseconds (`float64(time.Since(start))` in controller-runtime), but `metadata.csv` declared the unit as `millisecond`. The same raw value rendered as `1.45hr` instead of `5.24ms`, blocking latency SLOs/monitors. Fixed the unit to `nanosecond` for all five `reconcile_duration` metrics (`externalsecret`, `clusterexternalsecret`, `pushsecret`, `clustersecretstore`, `secretstore`).

### 2. Missing `controller_runtime.terminal_reconcile_errors`
Added the metric to the metric map (`metrics.py`) and `metadata.csv` (orientation `-1`), plus test fixture and expected-metric coverage.

### 3. Release
Bumped version `0.1.0` -> `0.1.1` and updated the CHANGELOG.

`sync_calls` peak discrepancy vs. custom metrics is expected scrape-interval alignment (not fixed in code).


willianccs:
@lucia-sb could you help here also? Thank you!

